PR #497, opened 2019-07-24
#496: exclude accidentally shaded artifacts

### LINKED ISSUE #496: 0.1.4 contains SLF4J-API classes
When importing factcast-core 0.1.4 via Maven the corresponding JAR file contains org.slf4j.MDC (and other classes from slf4j-api). Instead of providing the classes, I expect slf4j-api to be referenced using Maven.

uweschaefer:
closes #496

### DIFF
--- a/factcast-core/pom.xml
+++ b/factcast-core/pom.xml
@@ -38,6 +38,12 @@
               <goal>shade</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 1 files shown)


mweirauch:
Oops, didn't I mention artifactSet on the initial PR? Can you elaborate what went wrong?

Issue #601, opened 2020-01-09
Remove DDL/DML distinction as it causes more trouble than it is worth


uweschaefer:
see #548

PR #603, opened 2020-01-10
WIP #601: removed dml user mechanism

uweschaefer:


factcast-site/documentation/content/migration/_index.md
```diff
@@ -23,9 +23,10 @@ If you used the uniqe_identifier feature before, it was removed. It was only a r
 
 There is a [section on optimitic locking](/usage/java/optimistic_locking/) as a new api feature.
 
-#### Postgres module uuid-ossp and ddl user
+#### Postgres module uuid-ossp 
 
-The Postgres module *uuid-ossp* is necessary for the new optimistic locking api feature. In order to provide a user with enough privileges to install modules and also running the application without unnecessary db permissions, we add the possibility to distinct between _ddl user_ and _dml user_. This distinction is optional and your server will also work with your current database user configuration as long as the user is a superuser. If you don't want to provide a superuser and/or you want to distinct between _ddl_ and _dml user_, please take a look into the [Postgres section] (/setup/server/prerequisites/#postgres) in the prerequisites guide.
+The Postgres module *uuid-ossp* is necessary for the new optimistic locking api feature. In order to provide a user
```
mweirauch:
```suggestion
The Postgres module *uuid-ossp* is necessary for the new optimistic locking api feature. In order to install this
```

factcast-site/documentation/content/migration/_index.md
```diff
@@ -23,9 +23,10 @@ If you used the uniqe_identifier feature before, it was removed. It was only a r
...
+who is a superuser.
```
mweirauch:
```suggestion
extension, the user performing the Liquibase operations requires Postgres superuser permissions. 
```

factcast-store-pgsql/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -29,6 +29,3 @@ databaseChangeLog:
     - include:
          file: factcast/issue346/dropUniqueConstraintForIdent.sql
          relativeToChangelogFile: true
-    - include:
```
mweirauch: Is this backwards compatible to an existing 0.1.4.2 installation? Would have assumed the include is one changeset which then would be missing.

(Didn't try yet. Won't be able to test it today. But there's no hurry, right?)

uweschaefer: no problem there, changes already applied, but missing in changelog are just not taken into account.

uweschaefer: But please, test this change it in your context before i'll merge. no rush... it has time...

uweschaefer:
Reminder: Validation is needed here, if this also removes the inability to upgrade postgres to 11, due to granting access on the ser sequence. Otherwise, i'd create a different issue for that.

mweirauch:
Tested on a fresh postgres:9-alpine in the following sequence:
* run server with `0.1.4.2`
* run server with this branch
* add a temporary changeset and run server again based on this branch
* Result: liquibase ignores the missing/dropped changeset. So this seems to work.

Performing the same procedure against postgres:11-alpine works, but emitting facts results in the following issue, which likely mandates a separate issue. (I don't know any details about the problem from your previous comment.)

```
Caused by: org.postgresql.util.PSQLException: ERROR: null value in column "ser" violates not-null constraint
  Detail: Failing row contains (null, {"when": "2020-01-17T21:55:48.867+01:00", "message": "yeah"}, {"id": "47144689-6194-453d-a6c0-cb21fd6efad9", "ns": "testing-1"...).

```

uweschaefer:
could you recheck with pg 9.6 instead of 11?

mweirauch:
> could you recheck with pg 9.6 instead of 11?

The tests where performed against postgres:9.6.11-alpine and postgres:11-alpine (pointing to 11.6-alpine currently.)

uweschaefer:
thx, will have a look

uweschaefer:
now that is weird.

if you execute the current server against fresh postgres containers, liquibase behaves significantly different against 9.6.8 and 11.4:

 9.6.8:
```2020-01-19 10:29:55 [main] INFO  liquibase.executor.jvm.JdbcExecutor -
        CREATE TABLE public.fact (ser BIGSERIAL NOT NULL, payload JSONB NOT NULL, header JSONB NOT NULL, CONSTRAINT FACT_PKEY PRIMARY KEY (ser))
2020-01-19 10:29:55 [main] INFO  liquibase.changelog.ChangeSet - Table fact created
```

11.4:
```2020-01-19 10:28:47 [main] INFO  liquibase.executor.jvm.JdbcExecutor -
        CREATE TABLE public.fact (ser BIGINT NOT NULL, payload JSONB NOT NULL, header JSONB NOT NULL, CONSTRAINT FACT_PKEY PRIMARY KEY (ser))
2020-01-19 10:28:47 [main] INFO  liquibase.changelog.ChangeSet -
        Table fact created
```

that means there is no sequence for fact, as well as no default value (should have been ` nextval('fact_ser_seq'::regclass)`)

uweschaefer:
no idea why liquibase makes a BIGINT out of a BIGSERIAL when running against 11.

uweschaefer:
should also work with 11.4. please take another look.

uweschaefer:
closes #601

mweirauch:
So, to just get my testing observations clarified ...

postgres:9 automatically creates the sequence `fact_ser_seq` on initial DDL and also sets the default value on `fact.ser` correctly to `nextval('fact_ser_seq'::regclass)` whereas postgres:11 doesn't?

If so, possible to add some additional comments in the changeset sql?

Additionally, the `SELECT setval('fact_ser_seq', max(ser)+100) FROM fact;` is required for what kind of migration combination? In case from postgres:9 things should be fine as afterwards the defaults on `fact.ser` are still the same, even while inserting facts in parallel, no? Given the case one started with postgres:11 initially, one was never able to insert a fact, so incrementing `fact_ser_seq` is not necessary? Or I of course might just be missing something, so a helpful comment in the changelog sql would be helpful.

In any case, this shows that some migration tests should be written later migration combinations where liquibase changes are applied iteratively against different database versions. (Likely using testcontainers with shared volumes, if that is possible. So not just testing plain liquibase initialisations of the whole changeset, but switching db version in between.)

uweschaefer:
> postgres:9 automatically creates the sequence fact_ser_seq on initial DDL and also sets the default value on fact.ser correctly to nextval('fact_ser_seq'::regclass) whereas postgres:11 doesn't?

it is liquibase, rather than postgres. when talking to pg11, liquibase decides to use a BIGINT instead of a BIGSERIAL, event though BIGSERIAL is defined.

> Additionally, the SELECT setval('fact_ser_seq', max(ser)+100) FROM fact; is required for what kind of migration combination? 

i currently cannot come up with any. it is just a precaution. theroretically, when migrating data to a new engine, the schema is complete, so nothing needs to happen.

>  Given the case one started with postgres:11 initially, one was never able to insert a fact, so incrementing fact_ser_seq is not necessary?

true

i'll add some comments.

uweschaefer:
thx

mweirauch:
Oh, apparently I was of the assumption that `SELECT setval('fact_ser_seq', max(ser)+100) FROM fact;` would *not* make it into the merge. Now we are getting a non-contiguous sequence. (I know, highly nitpicky and perhaps irrelevant?)

uweschaefer:
as long as the seq a monotonous, everything is fine...

PR #617, opened 2020-01-17
Issue614 removed hugo binaries

uweschaefer:


factcast-site/documentation/serve.sh
```diff
@@ -16,6 +16,8 @@ rm -rf themes
 #)
 
 (
- ./hugo server --watch 
+docker run --rm -u `id -u` -v $PWD:/srv/hugo yanqd0/hugo hugo
```
mweirauch: Is this file supposed to serve for development purposes? Then why not having hugo permanently watch and serve from the container? This would get rid of the requirement for python locally.

    docker run --rm -u `id -u` -p 1313:1313 -v $PWD:/srv/hugo yanqd0/hugo hugo server --watch --bind 0.0.0.0

PR #619, opened 2020-01-18
#618: simplified spotless config

uweschaefer:


uweschaefer:
> The file .factcast-root should be dropped then I guess.

yep. i forgot, thx.

> mvn spotless:check -pl <some-module> doesn't detect issues neither on master nor on this branch.

yes, that is a choice. for performance reasons, spotless should only work on the reactor, not on the modules.

PR #662, opened 2020-02-17
Issue661 update jackson dependency

uweschaefer:


uweschaefer:
branched from an old master...

Issue #712, opened 2020-03-17
SchemaStore needs to depend on SpringLiquibase


uweschaefer:
in order to be sure liquibase ran before initializing the Store

PR #713, opened 2020-03-17
#712: declared dependency on liquibase

uweschaefer:


uweschaefer:
oh my. output folder not excluded...

PR #714, opened 2020-03-17
#712: declared dependency on liquibase

uweschaefer:


uweschaefer:
closes #712

uweschaefer:
@otbe thx

PR #720, opened 2020-03-22
WIP #715-factspec: added version to FactSpec

uweschaefer:
created convenient method to create FactSpec

### DIFF
--- a/factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
+++ b/factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
@@ -43,6 +43,9 @@ public class FactSpec {
     String type = null;
 
     @JsonProperty
+    Integer version = null;
+
+    @JsonProperty
     UUID aggId = null;
 
     @JsonProperty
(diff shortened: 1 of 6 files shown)


factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
```diff
@@ -42,6 +42,9 @@
     @JsonProperty
     String type = null;
 
+    @JsonProperty
+    Integer version = null;
```
uweschaefer: can be an int ? In FcatSpec::from, we declare 0 == undefined.

factcast-core/src/main/java/org/factcast/core/spec/FactSpecMatcher.java
```diff
@@ -102,6 +106,14 @@ protected boolean typeMatch(Fact t) {
         return type.equals(otherType);
     }
 
+    protected boolean versionMatch(Fact t) {
+        if (version == null) {
+            return true;
+        }
+        Integer otherVersion = t.version();
```
uweschaefer: else

factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
```diff
@@ -116,10 +116,10 @@ else if (filterScript == null && jsFilterScript != null)
     }
 
     public static <T> FactSpec from(Class<T> clazz) {
-        RenameMe annotationSpec = clazz.getAnnotation(RenameMe.class);
+        Specification annotationSpec = clazz.getAnnotation(Specification.class);
 
         if (annotationSpec == null) {
-            throw new IllegalArgumentException("You must annotate your Fact with @RenameMe");
+            throw new IllegalArgumentException("You must annotate your Fact with @Specification");
```
uweschaefer: @"+Specification.class.getSimpleName());

uweschaefer: just in case, it is renamed some day :D

Issue #889, opened 2020-06-14
Limit Auto-Reconnection behavior and escalate to application layer early.


uweschaefer:
Currently, ReconnectingFactSubscriptionWrapper runs in a loop trying to reconnect to the factcast server on connections problems.
While that might look good from a robustness point of view, the application layer needs to know if he runs stale due to a longer connection interruption.

This is why, when reconnect does not work immediately, it should not be retried, but escalated.

uweschaefer:
closed by #890

PR #890, opened 2020-06-14
Issue #889: limit reconnection attempts (reset after a while)

uweschaefer:


uweschaefer:
TODO: Needs to be documented.

PR #935, opened 2020-07-04
Issue934 optimize startup time of factcast-docker

uweschaefer:


uweschaefer:
reduced docker container startup by 20-30%

PR #951, opened 2020-07-12
Issue950 add metrics to roundtrip testing

uweschaefer:


uweschaefer:
duplication is ok for now, i have to unify this at some point

PR #989, opened 2020-08-19
#942_hook_for_tx_handling: suggestion for a hook

uweschaefer:


uweschaefer:
@BernhardBln added test, ready for review

factcast-factus/src/main/java/org/factcast/factus/applier/DefaultEventApplier.java
```diff
@@ -310,6 +311,11 @@ private static boolean isEventHandlerMethod(Method m) {
                 createSingleParameterTransformer(m, null, type);
             }
 
+            // exclude MockitoMocks
```
BernhardBln, line 314: Why would you want to do that?

factcast-factus/src/test/java/org/factcast/factus/DefaultFactusTest.java
```diff
@@ -297,7 +309,41 @@ void batchAbortedWithException() {
...
+            ManagedProjection m = Mockito.spy(new SimpleProjection());
+            EventApplier<ManagedProjection> ea = Mockito.spy(new DefaultEventApplier<>(mock(
+                    EventSerializer.class), m));
+            when(ehFactory.create(m)).thenReturn(ea);
+            ArgumentCaptor<FactObserver> observer = ArgumentCaptor.forClass(FactObserver.class);
+
+            Fact f1 = Fact.builder().ns("test").type("FooEvent").build("{}");
+            Fact f2 = Fact.builder().ns("test").type("FooEvent").build("{}");
+
+            when(fc.subscribe(any(), observer.capture()))
```
BernhardBln, line 342: I don't think you can use a captor in when, only in verify

BernhardBln (changes requested):
In case the test with the ArgumentCaptor in when() is green, please find out why :D

factcast-factus/src/main/java/org/factcast/factus/applier/DefaultEventApplier.java
uweschaefer: Otherwise, it will complain that methods are duplicate.
The mechanics will find all methods of all non-static inner classes. Unfortunately, Mockito generates an inner class extending the outer (weird isn't it) and thus duplicating the method.

factcast-factus/src/test/java/org/factcast/factus/DefaultFactusTest.java
uweschaefer: not true.

`    interface TestFN extends Function<Integer,String>{};

    @Test
    public void testCapturingInBehaviorDefinition(){

        TestFN mock = mock(TestFN.class);
        ArgumentCaptor<Integer> intCap = ArgumentCaptor.forClass(Integer.class);

        when(mock.apply(intCap.capture())).thenReturn("mockedValue");

        String res = mock.apply(112);

        Assertions.assertEquals("mockedValue",res);
        Assertions.assertEquals(112,intCap.getValue());
    }`

BernhardBln: Oh, indeed, seeing this for the first time... although not recommended :P 

"Warning: it is recommended to use ArgumentCaptor with verification but not with stubbing. Using ArgumentCaptor with stubbing may decrease test readability because captor is created outside of assert (aka verify or 'then') block. Also it may reduce defect localization because if stubbed method was not called then no argument is captured."

Well fine then.

PR #990, opened 2020-08-19
Micrometer integration with Factus

edthamm:


factcast-itests/factcast-itests-factus/src/main/java/org/factcast/itests/factus/Application.java
```diff
@@ -30,6 +32,7 @@
 @SuppressWarnings("ALL")
 @SpringBootApplication
 @Slf4j
+@EnableAutoConfiguration(exclude = CompositeMeterRegistryAutoConfiguration.class)
```
uweschaefer, line 35: why? i assume there is a story behind it?

uweschaefer (approved):
splendid! thank you guys...

PR #992, opened 2020-08-20
#942_suggestion_namespace_per_test: a suggestion to achieve namespace…

uweschaefer:
…-per-test behavior in order to be able to reuse eventclasses in more than one test without interference


uweschaefer:
@BernhardBln as discussed...

uweschaefer:
decided to drop that approach in favor of wiping PG after each integration test.

PR #1024, opened 2020-09-08
Issue1023 @ConditionalOnMissingBean on autoconfiguration

### LINKED ISSUE #1023: Factus autoconfiguration is always on
**Describe the bug**
`FactusAutoConfiguration` is always kicking in, not taking into account the already specified beans (contrarily to `FactCastAutoConfiguration`).

`0.3.0`

**To Reproduce**
Our tests declare a bean
```
  @Bean
  @Profile("test")
  public Factus factus() {
    return mock(Factus.class);
  }
```
in order to overwrite the autoconfiguration-provided Factus with a dummy implementation (a mocked one). As the autoconfiguration Factus bean doesn't have the `@ConditionalOnMissingBean` annotation, this fails the context launch.

**Expected behavior**
The autoconfigured Factus takes into account that _there might be something out there and maybe I don't have to jump in_ :)

**Additional context**
n/a

uweschaefer:
closes #1023


otbe (approved):
LGTM

PR #1050, opened 2020-09-21
JSON serialiser now calculates hash over schema

BernhardBln:


factcast-core/pom.xml
```diff
@@ -1,64 +1,71 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
uweschaefer, line 1: what is the change here?

factcast-core/src/main/java/org/factcast/core/snap/SnapshotId.java
```diff
@@ -29,4 +29,5 @@
     @NonNull
     UUID uuid; // semantically optional, might be constant if not needed
                // (SnapshotProjections)
+
```
uweschaefer, line 32: :grin:

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
```diff
@@ -158,10 +177,21 @@ public static JsonNode toJsonNode(Map<String, Object> jsonAsMap) {
...
         return objectMapper.readerFor(type).readValue(bytes);
-
     }
 
     public static ObjectMapper mapper() {
         return objectMapper;
     }
+
+    @SneakyThrows
+    public static long calculateHash(Class<?> projectionClass) {
```
uweschaefer, line 187: this uses the FCJson internal objectmapper, but we agreed that people must be able to swap the obejctmapper used for eventserialization. therefore i'd suggest to move this out of here and into the eventserializer ?

factcast-factus-bin-snapser/pom.xml
```diff
@@ -1,46 +1,53 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
uweschaefer, line 1: what is the change here?

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
```diff
@@ -158,10 +177,21 @@ public static JsonNode toJsonNode(Map<String, Object> jsonAsMap) {
...
+
+        JsonSchema jsonSchema = schemaGen.generateSchema(projectionClass);
```
uweschaefer, line 189: also i guess, the schema should be cached (Map<Type,Schema>)

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
```diff
@@ -158,10 +177,21 @@ public static JsonNode toJsonNode(Map<String, Object> jsonAsMap) {
...
+
+        String schema = writer
+                .writeValueAsString(jsonSchema);
+
+        return schemaModifier.apply(schema)
+                .hashCode();
```
uweschaefer, line 195: is the hashcode even guaranteed to be consistent? i'd expect taking the md5 of the string

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
uweschaefer: or better Ma<Type,Long> directly to the md5

uweschaefer (changes requested):
i'd suggest 

* MD5 instead of schema hash
* not recalculating the hash for a given class

factcast-core/pom.xml
BernhardBln: No clue, probably from spotless plugin?

factcast-core/src/main/java/org/factcast/core/snap/SnapshotId.java
BernhardBln: das sah aber auch aus :D

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
BernhardBln: Done in AbstractSnapshotRepository

factcast-core/src/main/java/org/factcast/core/util/FactCastJson.java
BernhardBln: changed

factcast-core/pom.xml
```diff
@@ -1,64 +1,71 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.11.2</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-jsonSchema</artifactId>
+            <version>2.11.2</version>
```
BernhardBln, line 38: How does it work with renovate-bot, would we remove the version here? Also above, in databind?

factcast-factus-bin-snapser/pom.xml
BernhardBln: added json-schema dependency

factcast-core/pom.xml
uweschaefer: i think there is a version defined in the properties. "jackson.version"?

BernhardBln:
@uweschaefer I cleaned up dependency management for jackson a bit:

- introduced property with version in root pom
- used this property for all jackson dependencies in bom
- removed version tag in all other poms

BernhardBln:
@uweschaefer what does that mean?

```
Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.4.1:check (default) on project factcast: 
     Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.4.1:check failed: 
          No such reference 'origin/master' -
```

https://github.com/factcast/factcast/pull/1050/checks?check_run_id=1175889328

uweschaefer:
crap. spotless is configured to change only files diffing compared to master. the message implies, that the repo is not fully cloned, but rather just a particular branch is sparse checked out. will look into it. thx

uweschaefer (approved):

uweschaefer:
thx @BernhardBln

Issue #1080, opened 2020-10-18
Move metric collection regarding snapshots to snapshotcache


uweschaefer:

uweschaefer:
closed by #1081

PR #1081, opened 2020-10-18
#1080 Move metric collection regarding snapshots to snapshotcache

uweschaefer:
**This PR Addresses:**  
[Move metric collection regarding snapshots to snapshotcache](https://github.com/factcast/factcast/issues/1080)


uweschaefer:
moved to snapshotrepositories, rather than snapshotcache, as the latter is an abstraction of factcast itself

otbe (approved):
LGTM

Issue #1075, opened 2020-10-15
Factus: When fetching SnapshotProjections takes a significant time, persist intermediate snapshots now and then in case another thread also fetches, or the fetching is interrupted before it ends.


uweschaefer:
for instance if fetching is part of a service startup, it take too long and the instance is being destoryed by a supervisor for not coming up quick enough (aws loadbalancer for instance).

Currently it would start from scratch and loop.

uweschaefer:
closed by #1082

PR #1082, opened 2020-10-18
#1075 Factus: When fetching SnapshotProjections takes a significant time, persist intermediate snapshots now and then in case another thread also fetches, or the fetching is interrupted before it ends.

uweschaefer:
**This PR Addresses:**  
[Factus: When fetching SnapshotProjections takes a significant time, persist intermediate snapshots now and then in case another thread also fetches, or the fetching is interrupted before it ends.](https://github.com/factcast/factcast/issues/1075)


otbe (approved):
LGTM

Issue #1084, opened 2020-10-19
Use an explicit getId() call instead of class.getCanonicalName() to avoid conflicts when dealing with anonymous classes.


uweschaefer:

uweschaefer:
closed by #1085

PR #1085, opened 2020-10-19
Use an explicit getId() call instead of class.getCanonicalName() to avoid conflicts when dealing with anonymous classes.

uweschaefer:
**This PR Addresses:**  
[Use an explicit getId() call instead of class.getCanonicalName() to avoid conflicts when dealing with anonymous classes.](https://github.com/factcast/factcast/issues/1084)


otbe (approved):
LGTM

PR #1091, opened 2020-11-04
Fixes #827 and improves other error message

### LINKED ISSUE #827: schema-registry-cli: Enrich validation result output.
### Problem
I get the following error when validating the current state of our schema registry:

```
$ java -jar fc-schema-cli.jar validate -p .
Starting validating Factcast Schema Registry
Input: /home/***/registry/src/main/registry

Warning: Nashorn engine is planned to be removed from a future JDK release
Example simple.json failed validation:
- ECMA 262 regex "^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$" does not match input string "TODO example missing"
- ECMA 262 regex "^[A-Z]{2}[0-9]{2}[0-9A-Z \t]{1,30}$" does not match input string "TODO example missing"
Example simple.json failed validation:
- ECMA 262 regex "^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$" does not match input string "TODO example missing"
- ECMA 262 regex "^[A-Z]{2}[0-9]{2}[0-9A-Z \t]{1,30}$" does not match input string "TODO example missing"

Validation failed!
```
It is now cumbersome to find out which event is missing a better example. 

### Suggestion
To relate to the actual event I believe it is enough to print out the full path of "simple.json".

Ideally, the line number and/or the errornous property is also printed.

samba2:
See commit comments for details

### DIFF
--- a/factcast-schema-registry-cli/src/main/kotlin/org/factcast/schema/registry/cli/validation/formatErrors.kt
+++ b/factcast-schema-registry-cli/src/main/kotlin/org/factcast/schema/registry/cli/validation/formatErrors.kt
@@ -31,9 +31,9 @@ fun formatErrors(errors: List<ProjectError>): List<String> = errors.map {
             "No description found for ${it.descriptionPath}"
         is ProjectError.ValidationError ->
             """
-Example ${it.examplePath.fileName} failed validation:
+Example ${it.examplePath} failed validation:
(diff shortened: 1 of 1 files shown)


uweschaefer (commented):
lgtm

Issue #1108, opened 2020-12-01
Check if lock is still held in subscribed projections


BernhardBln:
**Describe the bug**
Subscribed projections typically acquire a lock at application startup and release it only when shutting down.

This can turn into a problem when using remote locks like on Redis.

Maybe before updating the projection, check if the lock is still there (how to do this in a non-expensive way?) and try resubscribing if not.

uweschaefer:
In case of Redis this will be done with an occasional (not to limit throughput) lock.isLocked(). However, as Redis examples do not have a home in this repo yet, this issue will be closed and implemented where it is needed.
What needs to be done in FC however is the implementation of the necessary callbacks. see #1090

uweschaefer:
reopened to track the progress for the callback

PR #1124, opened 2020-12-08
#1108: lock liveness test

uweschaefer:


factcast-factus/src/main/java/org/factcast/factus/projection/WriterTokenAware.java
```diff
@@ -20,10 +20,10 @@
...
 
 public interface WriterTokenAware {
-  default AutoCloseable acquireWriteToken() {
+  default WriterToken acquireWriteToken() {
     return acquireWriteToken(FactusConstants.FOREVER);
   }
 
   /** might return null if token cannot be acquired */
-  AutoCloseable acquireWriteToken(@NonNull Duration maxWait);
+  WriterToken acquireWriteToken(@NonNull Duration maxWait);
```
BernhardBln, line 28: This would be an incompatible change that might break user code, right?

How about this solution (not very beautiful, but would not break anything):

```java
default AutoCloseable acquireWriteToken() {
  return acquireWriteToken(FactusConstants.FOREVER);
}

default AutoCloseable acquireWriteToken(@NonNull Duration maxWait){
  // note change in method name, writeToken -> writerToken
  throw new RuntimeException("Please implement acquireWriterToken!");
}

// note change in method name, writeToken -> writerToken
default WriterToken acquireWriterToken() {
  return acquireNewWriteToken(FactusConstants.FOREVER);
}

default WriterToken acquireWriterToken(@NonNull Duration maxWait){
   // If this is not overridden, assume we have legacy code overriding acquireWriteToken 
  AutoCloseable token = acquireWriteToken(maxWait);
  return new WriterToken() {

      @Override
      public void close() throws Exception {
        token.close();
      }

      @Override
      public boolean isValid() {
        return true;
      }
  };

}

```

And setting the old method deprecated and remove that sometime later?

factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/proj/RedisWriterToken.java
```diff
@@ -0,0 +1,56 @@
...
+    liveness = new AtomicBoolean(lock.isLocked());
+    val watchDogTimeout = redisson.getConfig().getLockWatchdogTimeout();
+    TimerTask timerTask =
+        new TimerTask() {
+          @Override
+          public void run() {
+            liveness.set(lock.isLocked());
+          }
+        };
+    timer.scheduleAtFixedRate(timerTask, 0, (long) (watchDogTimeout / 1.5));
```
BernhardBln, line 43: what is this magic with 1.5 doing here?

BernhardBln (commented):
I made a suggestion how to make this non-breaking.

factcast-factus/src/main/java/org/factcast/factus/projection/WriterTokenAware.java
uweschaefer: how is replacing an interface-return by an implementation return incompatible ?

factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/proj/RedisWriterToken.java
uweschaefer: 30 secs is default watchdog. i'd like to skip the test for 20 secs for performance reasons in order to be sure...

factcast-factus/src/main/java/org/factcast/factus/projection/WriterTokenAware.java
uweschaefer: this feels like trading a compile error for a runtime-error. where am i wrong?

BernhardBln: Yeah, for the sake of guaranteeing compatibility.

factcast-factus/src/main/java/org/factcast/factus/projection/LocalWriteToken.java
```diff
@@ -27,7 +27,7 @@
 
   private final ReentrantLock lock = new ReentrantLock(true);
 
-  public AutoCloseable acquireWriteToken(@NonNull Duration maxWait) {
+  public WriterToken acquireWriteToken(@NonNull Duration maxWait) {
     try {
       if (lock.tryLock(maxWait.toMillis(), TimeUnit.MILLISECONDS)) {
         return lock::unlock;
```
DanielGronau, line 33: I would suggest to overwrite `isValid`, else someone could reuse a lock by accident:
```
return new WriterToken() {
 @Override
  public void close() throws Exception {
    lock.unlock();
  }    

  @Overwrite
  public boolean isValid() {
    return lock.isLocked();
  }
}
```

DanielGronau (changes requested):

DanielGronau (approved):

PR #1126, opened 2020-12-08
Fixed NPE when updating projection

BernhardBln:
When updating a projection that is an anomymous class, we get a NullPointerException.


factcast-bom/pom.xml
```diff
@@ -1,130 +1,151 @@
...
+            <dependency>
+                <groupId>org.testcontainers</groupId>
+                <artifactId>testcontainers</artifactId>
+                <version>${testContainers.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.testcontainers</groupId>
+                <artifactId>postgresql</artifactId>
+                <version>${testContainers.version}</version>
+            </dependency>
```
BernhardBln, line 111: This is the only change I have done here, the rest is just formatting.

BernhardBln:
The diff is quite large due to spotles formatting of pom files.

Only changes done in BOM: added testContainers (and incremented version to 1.15.0)
Only changes done to POMs of sub modules: remove version for testContainers
Only change done to the root pom: 
        <etc>${maven.multiModuleProjectDirectory}/src/etc</etc>

uweschaefer:
closed in favor of #1133

Issue #1129, opened 2020-12-15
Factus: lock on a list of FactSpecs (as avail in factcast interface)


uweschaefer:
currently, the only way to lock on factus is to use projections, we face some usecases, where it would be necessary to lock on ajust a list of FactSpecs without updating anything.
Currently you'd have to use the factcast interface which has a slightly different way to express Attempts, which is confusing for Factus-only users.

uweschaefer:
Unit Tests are still missing after merge, keeping this ticket open.

PR #1130, opened 2020-12-15
#1129: Locking on specs only - some tests missing, but should be released asap

uweschaefer:


factcast-factus/src/main/java/org/factcast/factus/Factus.java
```diff
@@ -118,4 +119,10 @@ default void publish(@NonNull List<EventObject> eventPojos) {
 
   // conversion
   Fact toFact(@NonNull EventObject e);
+
+  LockedOnSpecs<? extends Projection> withLockOn(@NonNull FactSpec spec);
```
uweschaefer, line 123: the type parameter is pretty useless, should be cleaned up by replacing Locked by
LockedOnSpec and LockedOnProjection at some point...

otbe (approved):
LGTM👍

uweschaefer:
Ticket will be reopened after release for the purpose of adding appropriate unit tests

Issue #1107, opened 2020-12-01
Explore use of temp tables instead of catchup table


uweschaefer:
in order to reduce the size of the catchup table asap.

might become obsolete if we change the strategy (get rid of catchup table altogether)

uweschaefer:
this turned out to create too much extra load on the DB slowing us down in situations where high throughput is needed. we should explore temp tables here.

uweschaefer:
One idea is to use a SingleConnectionDataSource during catchup together with a session-local temp table with only the serials of the facts to return.
this way, there are no indexes to update and removal will happen automatically.

PR #1134, opened 2020-12-23
#1107: switched to temp table

uweschaefer:


factcast-store-pgsql/src/main/java/org/factcast/store/pgsql/internal/catchup/paged/PgPagedCatchup.java
```diff
@@ -50,60 +53,63 @@
...
+              } else {
+                log.trace("{} filtered id={}", request, factId);
               }
-            } else {
-              log.trace("{} filtered id={}", request, factId);
+            }
+
+            if (facts.isEmpty()) {
+              // we have reached the end
+              break;
```
DanielGronau, line 104: I think a do-while loop instead is more idiomatic

DanielGronau (approved):

factcast-store-pgsql/src/main/java/org/factcast/store/pgsql/internal/catchup/paged/PgPagedCatchup.java
uweschaefer: true

uweschaefer:
the global catchup table will be dropped via liquibase in the next major release.

DanielGronau (approved):

Issue #1128, opened 2020-12-10
Consider the impact of multimaps for header.meta


uweschaefer:
first idea is modifying the query (supposed to be compatible):

where header->>'meta.foo'= 'abc' OR header @> '{"meta.foo":["abc"]}';

but the impact on the query performance (if any) has to be investigated.

uweschaefer:
closed as won't fix until we come up with a good idea of how to do this in a compatible way

PR #1138, opened 2020-12-28
#1128: implement meta multimap

uweschaefer:


DanielGronau (approved):

uweschaefer:
closed as won't fix until we have a good idea how to do this in a compatible way.

Issue #1142, opened 2021-01-14
IllegalArgumentException with micrometer-registry-prometheus


Mortinke:
**Describe the bug**
We get this exception if we enable prometheus metrics:
```
java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'factcast_store_operations_seconds' containing tag keys [exception, operation, store]. The meter you are attempting to register has keys [operation, store].
	at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$collectorByName$9(PrometheusMeterRegistry.java:372)
	at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1877)
	at io.micrometer.prometheus.PrometheusMeterRegistry.collectorByName(PrometheusMeterRegistry.java:359)
	at io.micrometer.prometheus.PrometheusMeterRegistry.newTimer(PrometheusMeterRegistry.java:165)
	at io.micrometer.core.instrument.MeterRegistry.lambda$timer$2(MeterRegistry.java:271)
	at io.micrometer.core.instrument.MeterRegistry.getOrCreateMeter(MeterRegistry.java:580)
	at io.micrometer.core.instrument.MeterRegistry.registerMeterIfNecessary(MeterRegistry.java:543)
	at io.micrometer.core.instrument.MeterRegistry.timer(MeterRegistry.java:269)
	at io.micrometer.core.instrument.Timer$Builder.register(Timer.java:464)
	at org.factcast.store.pgsql.internal.PgMetrics.timer(PgMetrics.java:115)
	at org.factcast.store.pgsql.internal.listen.PgListener.checkDatabaseConnectionHealthy(PgListener.java:178)
	at org.factcast.store.pgsql.internal.listen.PgListener.receiveNotifications(PgListener.java:153)
	at org.factcast.store.pgsql.internal.listen.PgListener$NotificationReceiverLoop.run(PgListener.java:102)
	at java.lang.Thread.run(Thread.java:748)
```

**FactCast Server/Client Version**

0.3.8

**To Reproduce**

Add `micrometer-registry-prometheus` as dependency to enable prometheus metrics.
```
        <dependency>
            <groupId>io.micrometer</groupId>
            <artifactId>micrometer-registry-prometheus</artifactId>
        </dependency>
```

**Expected behavior**

We can start factcast with enabled prometheus metrics without exceptions.

PR #1143, opened 2021-01-14
#1142: ensure metrics have the same tags

Mortinke:


mweirauch:
Oopsie. This was apparently introduced when this was moved out of 'PgFactStore'. Would you be able to DRY/consolidate the `Tags.of()` construction into a single method like `private Tags forOperation(@NonNull PgMetrics.StoreMetrics.OP operation, @Nullable Exception e)` so that it's not redundant throughout the file? Currently there are two places which construct the tags.

So the `public Timer timer(OP)` could delegate to `private Timer timer(OP,ex)` which itself would make use of "forOperation()".

(All given the case I got this right by just looking at the code on GH.)

factcast-store-pgsql/src/main/java/org/factcast/store/pgsql/internal/PgMetrics.java
```diff
@@ -80,7 +81,7 @@ public void time(@NonNull OP operation, @NonNull Runnable r) {
     }
   }
 
-  public void time(@NonNull OP operation, @NonNull Sample sample, Exception e) {
+  private void time(@NonNull OP operation, @NonNull Sample sample, Exception e) {
```
uweschaefer, line 84: shouldn't this one stay public?

Mortinke: I don't see the use case to pass the start time for an operation. I only see this for distributed operations.

Issue #1166, opened 2021-02-02
incompatible dependencies grpc-*


StephanPraetsch:
**Describe the bug**
current version of factcast 0.3.9 seems to have misconfigured dependencies:
![image](https://user-images.githubusercontent.com/9827763/106670398-48cbde80-65ad-11eb-92da-5dabe96e4502.png)

It does not work with 1.33.0 stuff from grpc-* but with 1.31.1. 

1.33.0 causes
```text
Caused by: java.lang.NoSuchMethodError: 'io.grpc.internal.AbstractManagedChannelImplBuilder io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder.defaultLoadBalancingPolicy(java.lang.String)'
	at net.devh.boot.grpc.client.channelfactory.ShadedNettyChannelFactory.newChannelBuilder(ShadedNettyChannelFactory.java:73)
	at net.devh.boot.grpc.client.channelfactory.ShadedNettyChannelFactory.newChannelBuilder(ShadedNettyChannelFactory.java:50)
	at net.devh.boot.grpc.client.channelfactory.AbstractChannelFactory.newManagedChannel(AbstractChannelFactory.java:134)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
	at net.devh.boot.grpc.client.channelfactory.AbstractChannelFactory.createChannel(AbstractChannelFactory.java:104)
	at net.devh.boot.grpc.client.channelfactory.InProcessOrAlternativeChannelFactory.createChannel(InProcessOrAlternativeChannelFactory.java:83)
	at net.devh.boot.grpc.client.channelfactory.GrpcChannelFactory.createChannel(GrpcChannelFactory.java:72)
	at net.devh.boot.grpc.client.channelfactory.GrpcChannelFactory.createChannel(GrpcChannelFactory.java:51)
	at org.factcast.spring.boot.autoconfigure.client.grpc.GrpcFactStoreAutoConfiguration$1.createChannel(GrpcFactStoreAutoConfiguration.java:64)
	at org.factcast.client.grpc.GrpcFactStore.<init>(GrpcFactStore.java:94)
	at org.factcast.spring.boot.autoconfigure.client.grpc.GrpcFactStoreAutoConfiguration.factStore(GrpcFactStoreAutoConfiguration.java:72)
[...]

PR #1167, opened 2021-02-03
#1166: fixed grpc dependencies to 1.31.1

uweschaefer:


mweirauch:
Isn't it easier to import the `io.grpc:grpc-bom`? (haven't verified)

Btw, the `grpc.version` is contained in `factcast-internal-dep/pom.xml` ... perhaps it's correctable and or re-usable here? (don't have the project in an IDE in front of me)

PR #1172, opened 2021-02-09
Don't force the latest Maven 3.x patch release on the build

mweirauch:
Yes, there might be systems which lag one or two minors behind.

Additionally, it's apparently in the talks for a while to deprecate
the `prerequisites` section. Adding to that, the description of the
`maven`-element is somewhat misleading as it says something about
the Maven version the plugin is to be used with but actually a check
for the buid Maven version is performed.

I opted to `3.0`, but I guess dropping it entirely would also work.


uweschaefer:
fair enough

PR #1182, opened 2021-02-22
issue1180 Ignore transformations for blacklisted events

uweschaefer:


uweschaefer:
lgtm, will merge after pipeline has finished

PR #1184, opened 2021-02-24
issue1183: catch unchecked exception during catchup

uweschaefer:


factcast-store-pgsql/src/main/java/org/factcast/store/pgsql/internal/PgFactStream.java
```diff
@@ -94,16 +94,22 @@ private void catchupAndFollow(
       SubscriptionRequest request, SubscriptionImpl subscription, PgSynchronizedQuery query) {
     if (request.ephemeral()) {
       // just fast forward to the latest event publish by now
-      this.serial.set(fetcher.retrieveLatestSer());
+      serial.set(fetcher.retrieveLatestSer());
     } else {
-      catchup(postQueryMatcher);
+      try {
+        catchup(postQueryMatcher);
+      } catch (Throwable e) {
```
samba2, line 101: Did the exception we catch here appear somewhere higher in the call hierarchy? So did we see something suspicious before? Also, I assume we don't know what can go wrong here - or why is the catch clause so generic?

samba2 (approved):

Issue #1238, opened 2021-03-20
Replace use of lombok val (to help with compiler performance)


uweschaefer:

uweschaefer:
handled by #1463

PR #1271, opened 2021-04-13
#1238: removed val usage

uweschaefer:


uweschaefer:
closes #1238

uweschaefer:
supseeded by #1463

PR #1272, opened 2021-04-13
#1233: improved logging and readability (hopefully)

### LINKED ISSUE #1233: Improve logging when "channel not coming back"
**Describe the bug**
"channel does not come back" is currently logged in error level for different reasons:
* the client cancelled the subscription
* the client exceeds the max time to process messages

At least in the CANCELLED case, this should not be error level, and the message should clearly describe what was going on.

uweschaefer:
closes #1233

### DIFF
--- a/factcast-server-grpc/src/main/java/org/factcast/server/grpc/BlockingStreamObserver.java
+++ b/factcast-server-grpc/src/main/java/org/factcast/server/grpc/BlockingStreamObserver.java
@@ -85,8 +85,15 @@ public class BlockingStreamObserver<T> implements StreamObserver<T> {
           if (delegate.isReady()) {
             delegate.onNext(value);
           } else {
+
+            // no longer ready. has it been canceled in the meantime?
             if (!isCancelled(delegate)) {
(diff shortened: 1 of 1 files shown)


factcast-server-grpc/src/main/java/org/factcast/server/grpc/BlockingStreamObserver.java
```diff
@@ -61,29 +61,61 @@ void wakeup() {
...
-          if (!delegate.isReady() && !delegate.isCancelled()) {
-            throw new TransportLayerException("channel not coming back.");
+        }
+
+        // what about now?
+        if (!isCancelled(delegate)) {
+          if (delegate.isReady()) {
+            delegate.onNext(value);
+          } else {
+            if (!isCancelled(delegate)) {
```
otbe, line 88: Missing case ?

uweschaefer: nope. added some comments

otbe (approved):
LGTM (hopefully 😅)

PR #1284, opened 2021-04-16
issues1261 fix md5sum in index.json

samba2:


factcast-schema-registry-cli/src/main/kotlin/org/factcast/schema/registry/cli/registry/impl/IndexFileCalculatorImpl.kt
```diff
@@ -29,67 +30,82 @@ import org.factcast.schema.registry.cli.utils.mapEventTransformations
...
 
         return Index(schemas, transformations.plus(syntheticTransformations))
     }
+
+    private fun createMd5Hash(filePath: Path, schemaStripTitles: Boolean): String =
+            if (schemaStripTitles) createTitleFilteredMd5Hash(filePath)
+                else checksumService.createMd5Hash(filePath)
+
+    private fun createTitleFilteredMd5Hash(filePath: Path): String {
+        val filteredJsonNode = titleFilterService.filter(
```
uweschaefer, line 106: using a "service" here is probably overblown as it must not have state, and does not have dependencies either.
a simple function would do here.

uweschaefer: Also, TitleFilterServiceImpl deals with nullable, which it should not?

uweschaefer (approved):
lgtm

factcast-schema-registry-cli/src/main/kotlin/org/factcast/schema/registry/cli/registry/impl/IndexFileCalculatorImpl.kt
samba2: > using a "service" here is probably overblown as it must not have state, and does not have dependencies either.
> a simple function would do here.

I totally agree. For some reasons we've implemented it as a filter in the first place. You're right - sharing a stateless function can be done much easier.

uweschaefer: @samba2 do you want to commit the nullability issue?

PR #1285, opened 2021-04-18
Pom cleanups

mweirauch:
Some POM cleanups regarding jackson which hopefully reduce the amount of PRs and other integration runs on the next version bump if I didn't miss a spot.

Another build warning fix for testcontainers and moved slf4j dependency around also reducing the amount of touched files on version bump.

Let me know if you want these separate from each other.


uweschaefer (approved):
thx @mweirauch

PR #1300, opened 2021-04-29
Update DefaultProjector.java

Umang228:


uweschaefer:
your patch does not compile and does not solve the problem. sorry, i'll close this.

Issue #1161, opened 2021-02-02
Use TOAST storage in postgres for payloads


uweschaefer:
In order to speed up seq scanning, it'd be good to use TOAST storage in postgres for fact payloads.

This will probably involve adding a liquibase patch. Note however, that changing the setting only works for future rows, so that we should see how to make this change for existing rows, and if this possible to do online, and what it does to the load of the database.

uweschaefer:
detail about toast:
https://wiki.postgresql.org/wiki/TOAST

PR #1317, opened 2021-05-14
Issue1161 adjust TOAST settings

uweschaefer:
and cleanup some places where json was stored as text


uweschaefer:
closes #1161

factcast-store-pgsql/src/main/resources/db/changelog/factcast/issue1161/storage-setting.sql
```diff
@@ -0,0 +1,19 @@
+alter table fact
+    alter column header set storage MAIN;
+alter table fact
+    alter column payload set storage EXTENDED;
```
samba2, line 4: Interesting, but I don't fully get it. The doc says on high-level _A table that has columns with potentially large entries will have an associated TOAST table, which is used for out-of-line storage of field values that are too large to keep in the table rows proper._ I would not consider an average fact a large entry - what benefit does using TOAST bring for FactCast?

samba2 (approved):
lgtm, one question open

factcast-store-pgsql/src/main/resources/db/changelog/factcast/issue1161/storage-setting.sql
uweschaefer: when scanning, it is important to have the header inline with the row in case of filtering, whereas - if a row should exceed 8k - it should be the payload being stored out of band, not the header.
The sole purpose of this is i/o cache efficiency.

PR #1333, opened 2021-05-28
Fix transform handling when adding a new array

olewehrmeyer:
Currently Factcast includes a bug, where event transformations that add a new array are handled incorrectly.

When using a transformation function like this:
```js
function transform(event) {
  event.newField = [1 ,2 ,3];
}
```
the new field would be transformed to
```json
{
  "newField" : {
    "0": 1,
    "1": 2,
    "2": 3
  }
}
```

This happens as Jackson mishandles the Nashorn internal `ScriptObjectMirror`, which is an array if `ScriptObjectMirror#isArray == true` and an object otherwise. As it however implements `Map` internally, Jackson treats it as a Map.

---

Implementation note: I decided against telling Jackson how to treat this `ScriptObjectMirror` as I for one did not want to touch the `FactCastJson` class due to the "do not change mapper" comment and it kind of felt wrong to configure Jackson how to deal with this Nashorn weirdness, so I opted to simply filter this before passing it to Jackson.


uweschaefer:
merging it to a repo internal branch

Issue #1350, opened 2021-06-10
Make FC startable, even if schemareg update fails


uweschaefer:
and instead rely on scheduled update

PR #1376, opened 2021-06-22
#1350: tolerate schema reg update issue on startup

uweschaefer:


uweschaefer:
closes #1350

factcast-store-pgsql/src/main/java/org/factcast/store/pgsql/registry/AbstractSchemaRegistry.java
```diff
@@ -87,8 +88,14 @@ public void fetchInitial() {
     synchronized (mutex) {
       Stopwatch sw = Stopwatch.createStarted();
       log.info("Registry update started");
-      indexFetcher.fetchIndex().ifPresent(this::process);
-      log.info("Registry update finished in {}ms", sw.stop().elapsed(TimeUnit.MILLISECONDS));
+      try {
+        indexFetcher.fetchIndex().ifPresent(this::process);
+        log.info("Registry update finished in {}ms", sw.stop().elapsed(TimeUnit.MILLISECONDS));
+      } catch (Throwable e) {
+        log.warn("While fetching initial state of registry", e);
```
otbe, line 95: I wonder if we should handle this differently depending on how `persistentSchemaRegistry` is set. 🤔
What I mean is if theres no persistent schema registry booting an empty FC also makes no sense right? 🤔

uweschaefer: well. that was the initial assumption, which turned out to be impractical, hence this ticket... :thinking:

otbe (approved):

Issue #1388, opened 2021-06-24
Transport client "name" as part of request metadata


uweschaefer:
useful for:
* DOS protectoin (buckets)
* Logging
* client version matrix

PR #1396, opened 2021-06-27
issue1388: added client.id to grpc metadata & added in loglines

uweschaefer:
closing #1388

### DIFF
--- a/factcast-server-grpc/src/main/java/org/factcast/server/grpc/GrpcRequestMetadata.java
+++ b/factcast-server-grpc/src/main/java/org/factcast/server/grpc/GrpcRequestMetadata.java
@@ -51,7 +51,7 @@ public class GrpcRequestMetadata {
   public Optional<String> consumerId() {
     Optional<String> ret = Optional.empty();
     if (headers != null) {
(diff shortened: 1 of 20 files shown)


factcast-server-grpc/src/main/java/org/factcast/server/grpc/GrpcRequestMetadata.java
```diff
@@ -45,4 +47,12 @@ public static GrpcRequestMetadata forTest() {
     grpcRequestMetadata.headers.put(Headers.FAST_FORWARD, "true");
     return grpcRequestMetadata;
   }
+
+  public Optional<String> clientId() {
+    Optional<String> ret = Optional.empty();
```
otbe: Maybe use map/flatMap here?

otbe (approved):
LGTM, small aesthetic comment :D

factcast-server-grpc/src/main/java/org/factcast/server/grpc/GrpcRequestMetadata.java
uweschaefer: was changed already

Issue #1456, opened 2021-07-27
Fix default SpringTransactional timeout


samba2:
Docs of `DefaultTransactionDefinition.setTimeout(...)`: Set the timeout to apply, as number of seconds.

otbe:
please use ms as base

uweschaefer:
decided for seconds even though the granularity is not optimal for tests, because otherwise we'd need to round when using springTX, as the timeout is only declarable in seconds there

PR #1457, opened 2021-07-27
#1456 fix spring transactional timeout

samba2:


samba2:
see https://github.com/factcast/factcast/issues/1456

Issue #1421, opened 2021-07-14
Introduce auto-resilient subscriptions


uweschaefer:
**Is your feature request related to a problem? Please describe.**
Sometimes (with unclear cause) the server terminates a subscription (channel gone or similar) without the client noticing.
In this case data gets stale until resubscription, which obviously is a problem when you cannot know, when a resubscription is needed.

What already exists:

* If the consumer detects a networking issue, it automatically resubscribes which proved to work fine.
* The consumer is supposed to send protocol level keep-alives (GRPC setting) which *should* already detect broken subscriptions, but this does not seem to work.

**Describe the solution you'd like**
We'd like to be able to configure subscription in a way that actively detects abandoned subscriptions from the client side.
This could be done by implementing an application-level keepalive that is sent by the server (if asked for it) after N seconds of silence on the wire in order to signal _dont't worry, i am still here_.

If the client detects a silence period far beyond what was agreed, it would be his responsibility to close and reconnect in order to be sure it always has _not too stale_ data.

If this _kind of 'keepalive'_ interval can be configured from the client, it should have a lower bound in order not to stress the server too much.

Also, the application should be able to recieve callbacks for reconnections etc if it so chooses.

PR #1461, opened 2021-07-28
#1421 applevel keepalive

uweschaefer:


factcast-core/src/main/java/org/factcast/core/subscription/SubscriptionRequestTO.java
```diff
@@ -42,7 +42,9 @@
 
   private static final String PID = ManagementFactory.getRuntimeMXBean().getName();
 
-  @JsonProperty long maxBatchDelayInMs = 0;
+  @JsonProperty long maxBatchDelayInMs = 20;
```
otbe, line 45: 10 or 20? 🤔

factcast-itests/factcast-itests-exception-handling/src/test/java/org/factcast/itests/exception/handling/TransportLayerExceptionHandling.java
```diff
@@ -0,0 +1,97 @@
...
+          public void onError(@NonNull Throwable exception) {
+            System.out.println("onErr " + exception);
+          }
+        };
+    fc.subscribe(
+            SubscriptionRequest.follow(Lists.newArrayList(FactSpec.ns("users").type("UserCreated")))
+                .fromScratch(),
+            obs)
+        .awaitCatchup();
+    System.out.println("cuuuuuuuu");
```
otbe, line 85: Should this stay here? 🤔

factcast-core/src/main/java/org/factcast/core/subscription/SubscriptionRequestTO.java
uweschaefer: tought we can go for 20 as a default (concerned about load on database, rather than 10ms more latency)
what do you think?

factcast-itests/factcast-itests-exception-handling/src/test/java/org/factcast/itests/exception/handling/TransportLayerExceptionHandling.java
uweschaefer: certainly not :older_man:

Issue #1462, opened 2021-07-29
Add alternative ScriptEngine 


otbe:
Because of the deprecation of nashorn in java11 we should add another execution environment for our tranformations.  For example https://golb.hplar.ch/2020/04/java-javascript-engine.html

PR #1468, opened 2021-07-30
#1462: Introduce GraalJs based ScriptEngine 

otbe:
As preparation for java11 we introduce a GraalJs based Transfomer.


uweschaefer (approved):
lgtm

PR #1471, opened 2021-07-30
Docs + example code for issues1148 - Transactional Projections & Spring Transactional Projections

samba2:
best viewed with:
`cd factcast-site/documentation/ && ./serve.sh`

http://localhost:1313/factus/transactional_projections/

PR #1475, opened 2021-08-02
issue1147: docs for RedisTransactional

samba2:
This MR builds up on https://github.com/factcast/factcast/pull/1471  (Spring Transactional)
To enable linking via writing, it also contains some placeholders which must be replaced by the other MRs version.

Note: this is still draft as we should discuss the common configuration. Currently we say in the introduction to "Transactional Projections" that these 2 parameters are common between all three transactional projection  types:
- bulkSize
- timeoutInSeconds

However, `@RedisTransactional` and `@RedisBatched` have a `size` and `timeout` element.


uweschaefer:
included in #1485

Issue #1463, opened 2021-07-29
Upgrade to java11


otbe:
- update for compilation & runtime
- client libraries should be tested against 8 & 11

uweschaefer:
recheck that dependency tree does not force j11 on client

BernhardBln:
Check done, some modules were forced down to 1.8

uweschaefer:
closed by #1479

PR #1479, opened 2021-08-02
Issue1463 move to java11, disallow use of lombok.val/var

uweschaefer:


uweschaefer:
closes #1463

uweschaefer:
due to the crosscutting nature of the change, the coverage change is irrelevant.

PR #1489, opened 2021-08-05
Update image link in projections.md

h2000:
The projection.png is missing, could not find a better one / or the sources for the image, replaced it with the ph_p.png .


uweschaefer (approved):
exactly right - thx @h2000

PR #1513, opened 2021-08-20
issue1510 ffwd on non-empty catchups

uweschaefer:


factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -164,15 +167,11 @@ void fastForward(SubscriptionRequest request, SubscriptionImpl subscription) {
         startedSer = idToSerMapper.retrieve(startedId); // should be cached anyway
       }
 
-      if (!factsHaveBeenSent(startedSer, serial)) {
-        // we have not sent any fact. check for ffwding
+      UUID targetId = ffwdTarget.targetId();
```
otbe, line 170: Is `factsHaveBeenSent` still needed? 🤔

otbe (approved):
LGTM

factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
uweschaefer: u right

PR #1529, opened 2021-08-23
issue1522 swallow and transmit server side exception also in half-closed state

uweschaefer:


uweschaefer:
coverage may be ignored here :older_man:

otbe (approved):
LGTM

PR #1604, opened 2021-09-22
issue1595: factus: handlers: filtering by annotation

uweschaefer:


samba2:
doc: added some links and tried to make some statements more explicit. Pls double check

samba2:
was just thinking. I see three ways of filtering which I would like to have contrasted (maybe at the filtering.md) page:
- server-side via new annotations (static)
- server-side via `postprocess` (dynamic, can be also in combination with annotations)
- client-side via handling the event and returning if a predicate is not matched (early return)

should we add this here?

uweschaefer:
thx. i don't think we should advertise client side filtering at all, and usage of postprocess and these annotations is properly documented, imho

samba2:
lgtm

samba2 (commented):
lgtm

samba2 (approved):
lgtm

PR #1635, opened 2021-10-06
issue1634 tools using threadlocal fail due to GRPC using different thread for subsequent calls to onNext  

uweschaefer:
see `factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/GrpcThreadingITest.java`


uweschaefer:
closes #1634

Issue #1590, opened 2021-09-16
Integration Test for notification trigger extension


uweschaefer:
**Describe the bug**
the notification coming from fact inserts should contain a TXID. There needs to be an integration test for this.

BernhardBln:
Merged; will create a new ticket regarding the issue mentioned here: https://github.com/factcast/factcast/pull/1657#issue-1035961324

PR #1657, opened 2021-10-26
#1590-integration-test-for-notification-trigger-extension

BernhardBln:
This now tests that the tx id is part of the trigger, and is send around in the event on the event bus.

Note: the integration tests in factcast-store do not use the official sql, but a test schema creation file, so changes in the official sql would not turn this test red. My suggestion, however, is to consider fixing this in general, rather than making an exception for this specific test. I'm unaware of the reasons for this separation though.

@uweschaefer any thoughts?


samba2 (approved):
lgtm. tnx for learning about "anySatisfy"

PR #1663, opened 2021-10-27
Issue1162 harden grpc layer to enforce security constraints

### LINKED ISSUE #1162: Harden gRPC layer to enforce security constraints
The gRPC server layer needs some review in order to make sure all operations are properly authorized. ~~Also the authorization configuration should be extended in order to not only define who can read/write to what namespace, but also to have some means of access management to more technical/administrative endpoints.~~

BernhardBln:
Fixes #1162 

✅ Did harden:

- stateFor
- stateForSpecsJson

⚠️ Did not harden those:

- serialOf: no check, but I guess is fine to get the serial of an event by uuid even if this event is in a ns I shouldn't be able to access

- get/set/clear snapshots: difficult, would not know how to do this on the server side...

### DIFF
--- a/factcast-client-grpc/pom.xml
+++ b/factcast-client-grpc/pom.xml
@@ -99,4 +99,24 @@
       <optional>true</optional>
     </dependency>
   </dependencies>
+  <build>
+    <resources>
+      <!-- do not filter src/main/resources, except for version.properties,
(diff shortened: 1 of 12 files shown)


uweschaefer:
cherry picked a few things, rest is obsolete

PR #1668, opened 2021-11-01
#1633-make-snapshotserializersupplier-take-a-list-of-snapshotserializ…

### LINKED ISSUE #1633: Make SnapshotSerializerSupplier take a list of SnapshotSerializer's to be able to use beans with non default constructor
Right now we are only able to use
```
@SerializeUsing(SnapSer.class)
```

with classes that have a no-arg constructor but this is not strictly necessary 🤔

BernhardBln:
fixes #1633

### DIFF
--- a/factcast-factus/src/main/java/org/factcast/factus/snapshot/SerializeUsing.java
+++ b/factcast-factus/src/main/java/org/factcast/factus/snapshot/SerializeUsing.java
@@ -24,5 +24,5 @@ import org.factcast.factus.serializer.SnapshotSerializer;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 public @interface SerializeUsing {
-  Class<? extends SnapshotSerializer> value();
+  Class<? extends SnapshotSerializer>[] value();
 }
--- a/factcast-factus/src/main/java/org/factcast/factus/snapshot/SerializerInstantiationException.java
+++ b/factcast-factus/src/main/java/org/factcast/factus/snapshot/SerializerInstantiationException.java
@@ -19,4 +19,7 @@ public class SerializerInstantiationException extends RuntimeException {
   public SerializerInstantiationException(String msg, Throwable e) {
     super(msg, e);
   }
+  public SerializerInstantiationException(String msg){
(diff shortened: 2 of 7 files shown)


BernhardBln:
Unfortunately, this is _kind of_ a breaking change:

The change suggests that we will have now more than one bean.

The component however expects one bean to be the default serializer. Before, it was simply _the_ `SnapshotSerializer` bean that was in the container.

Now (after adding further beans), it has to be a bean with the name `snapshotSerializer`.

Before, it would work if a bean with any name was configured, as long as the type was `SnapshotSerializer`.

**However, this will not break anything when just updating the dependency, only when starting to declare further serializer beans:**

There was either only one `SnapshotSerializer` bean, or if there were more than one, there must have been one `SnapshotSerializer` with name `snapshotSerializer`, otherwise the application would not have come up.

So basically, when updating and adding further serializer beans, one needs to rename the previously default serializer to `snapshotSerializer`, if it has not been the case.

BernhardBln:
Regarding this comment (https://github.com/factcast/factcast/pull/1668#issuecomment-956232601), we decided to make the following change:

The expected name of the default snapshot serializer bean has to be `defaultSnapshotSerializer`. 

If you have not declared your own serializer, or you only declared one (which we believe should cover 99% of all cases), you can safely update, and you only run into problems when you start declaring further serializers (then you need to make sure that the default one is called `defaultSnapshotSerializer`).

If you have already more than one, then you already must have called the default one `ser`, as otherwise, the application would not have come up. In  that case, you would have to rename that one to `defaultSnapshotSerializer`.

Since this is still not 100% clean, we decided not to proceed here until we get an OK from @uweschaefer.

(Then we also still need an MR)

uweschaefer:
Just wondering: As serializers are supposed to be stateless and just turn Projections into bytes, what is the actual usecase here?

BernhardBln:
> Just wondering: As serializers are supposed to be stateless and just turn Projections into bytes, what is the actual usecase here?

See description in ticket: https://github.com/factcast/factcast/issues/1633#issue-1018258779

uweschaefer:
superseeded by #2608

PR #1669, opened 2021-11-02
Issue1170 add administrative endpoint for counting facts according to factspecs

### LINKED ISSUE #1170: Add administrative endpoint for counting facts according to factspecs
This is useful in potential UIs (for drilldown filtering for example), for administrative purposes (check if mass publishing went as expected) and maybe e2e tests involving mass data.

However, counting in PG is very expensive and we should try to make sure, this feature is not abused for programmatic puroposes.

BernhardBln:
Fixes #1170

### DIFF
--- a/factcast-client-grpc/pom.xml
+++ b/factcast-client-grpc/pom.xml
@@ -99,4 +99,24 @@
       <optional>true</optional>
     </dependency>
   </dependencies>
+  <build>
+    <resources>
+      <!-- do not filter src/main/resources, except for version.properties,
+        where we want the project version to be replaced -->
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>false</filtering>
+        <excludes>
+          <exclude>version.properties</exclude>
+        </excludes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>version.properties</include>
+        </includes>
+      </resource>
+    </resources>
+  </build>
 </project>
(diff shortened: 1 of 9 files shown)


factcast-server-admin-ui/src/main/java/org/factcast/server/admin/ui/controller/CountingController.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.server.ResponseStatusException;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Controller
+@RequestMapping("/admin/counting")
+public class CountingController {
```
uweschaefer, line 24: with administrative endpoint, we meant a service method in grpc next to enumerateNamespaces.
also we'd want to not technically run a select count(*), as it would be a DOS vector.
Security is not an issue, we'd need to add a security role for administrative access, where all administrative methods can be called

BernhardBln: Ok. If I remember correctly there was another ticket about a factcast UI, should I park this on the branch? For now, this is just a simple POC for a UI.
So how else you want to implement this, if not with a count?

uweschaefer: yes, please leave UI out of it here.
There are some alternatives to postgres count() to be found on the web, which are not guaranteed to be exact. please test them (with a reasonably huge fact table) and pick the best suited one.

Issue #1696, opened 2021-12-03
Increase gin_pending_list_limit


uweschaefer:
for the non-partial GIN index, as well as the tail indexes.
The default is 4m that might be too small, at least in batch insert scenarios.

PR #1772, opened 2022-01-28
issue1696 added setting for gin fast update

uweschaefer:


uweschaefer:
fixes #1696

samba2:
lgtm

Tnx for letting me review that. I read the postgres docs so I understand the speed improvement comes from *...the entries are moved to the main GIN data structure using the same bulk insert techniques used during initial index creation. This greatly improves GIN index update speed,...* [(link)](https://www.postgresql.org/docs/11/gin-implementation.html#GIN-FAST-UPDATE)

Did you do some comparison (before/ after) ? I understand that the default of 4 MB for the gin_pending_list was increased to 32 MB. Would be interesting to see the effect of this.

samba2 (approved):
lgtm, see other comment

uweschaefer:
Nope, it is just a precaution for a situation where we we're bulk ingesting.

Issue #1694, opened 2021-11-30
Enable parallel validation of facts on publish


uweschaefer:
We want to be as fast as possible accepting huge batches of facts (migration scenario) so the validation for say 1000 facts would benefit from parallelization.

PR #1775, opened 2022-01-28
Issue1694 parallel validation

uweschaefer:


uweschaefer:
closes #1694

factcast-store/src/main/java/org/factcast/store/registry/validation/FactValidationAspect.java
```diff
@@ -46,10 +47,9 @@ public Object interceptPublish(ProceedingJoinPoint joinPoint) throws Throwable {
   }
 
   private void validate(List<? extends Fact> facts) {
-
```
otbe, line 49: dunno but maybe we should use a dedicated thread pool here instead of the common one?
Just to ensure that "validation" load is an isolated thing

uweschaefer: done

otbe (approved):
LGTM

PR #1782, opened 2022-02-02
Allow to remove arbitrary fields from schema jsons before calculatin hash

### LINKED ISSUE #1781: Allow to remove arbitrary fields from schema jsons before hash cacualtion
It would be nice to be able to not only strip the "title" off from the schema files before calculating the hash. People can add for example "description" or "example" fields in there and they would cause the same issue as "title" for our hash calculation.

otbe:
fixes #1781 

You can add `--schema-remove-fields title,description,example` to the cli call or via maven plugin

```
<removeSchemaFields>
   <removeSchemaField>title</removeSchemaFields>
   <removeSchemaField>description</removeSchemaFields>
   <removeSchemaField>example</removeSchemaFields>
</removeSchemaFields>
```

The current implementation is backwards compatible and allows to mix the new stuff with `schemaStripTitles`


uweschaefer (approved):
fine fine

PR #1824, opened 2022-02-21
Issue1823 reduce work needed on optimistic locking

uweschaefer:


uweschaefer (commented):
nice, thx @bresciamattia

Issue #1865, opened 2022-03-24
Fix index for enumeration


uweschaefer:
`"index_for_enum" btree ((header ->> 'ns'::text), (header -> 'type'::text))`
should be 
`"index_for_enum" btree ((header ->> 'ns'::text), (header ->> 'type'::text))
`

PR #1870, opened 2022-03-29
#1865: add liquibase changeset to fix index for enumeration

bresciamattia:


factcast-store/src/main/resources/db/changelog/factcast/issue1865/fix_index_for_enum.sql
```diff
@@ -0,0 +1,3 @@
+-- fix the index index_for_enum from issue170
+drop index if exists index_for_enum;
+create index index_for_enum on fact using btree((header->>'ns'),(header->>'type'));
```
uweschaefer, line 3: the new index needs to be created concurrently in order not to lock the fact table

PR #1908, opened 2022-04-24
Issue1898 Refactor error / retry handling 

uweschaefer:


factcast-client-grpc/src/main/java/org/factcast/client/grpc/ResilientGrpcSubscription.java
```diff
@@ -0,0 +1,298 @@
...
+    public void onFastForward(@NonNull UUID factIdToFfwdTo) {
+      originalObserver.onFastForward(factIdToFfwdTo);
+    }
+
+    @Override
+    public void onError(@NonNull Throwable exception) {
+      log.info("Closing subscription due to onError triggered.  ({})", originalRequest, exception);
+      closeAndDetachSubscription();
+
+      resilience.registerAttempt();
```
otbe, line 221: Shouldnt this be inside the if? 🤔

uweschaefer: depends on how you interpret resilience.retries = 10
is this going to be fail after 10 failures or after 11?

Thinking about it, should we rename the property to attempts ?

Issue #1681, opened 2021-11-15
Have descriptive error message when publishing fails due to duplicate id


otbe:
Under some heavy load situations it seems like that we're able to publish events even if the FC returns an exception to the client.

For example, on the client side we get:

```
io.grpc.StatusRuntimeException: UNAVAILABLE: unavailable
```

while the FC logs
```
java.lang.IllegalStateException: call already closed
	at com.google.common.base.Preconditions.checkState(Preconditions.java:508)
	at io.grpc.internal.ServerCallImpl.closeInternal(ServerCallImpl.java:209)
	at io.grpc.internal.ServerCallImpl.close(ServerCallImpl.java:202)
	at io.opentelemetry.javaagent.shaded.instrumentation.grpc.v1_6.TracingServerInterceptor$TracingServerCall.close(TracingServerInterceptor.java:98)
	at io.grpc.PartialForwardingServerCall.close(PartialForwardingServerCall.java:48)
	at io.grpc.ForwardingServerCall.close(ForwardingServerCall.java:22)
	at io.grpc.ForwardingServerCall$SimpleForwardingServerCall.close(ForwardingServerCall.java:39)
	at net.devh.boot.grpc.server.metric.MetricCollectingServerCall.close(MetricCollectingServerCall.java:60)
	at org.factcast.server.grpc.GrpcServerExceptionInterceptor$ExceptionHandlingServerCallListener.handleException(GrpcServerExceptionInterceptor.java:104)
	at org.factcast.server.grpc.GrpcServerExceptionInterceptor$ExceptionHandlingServerCallListener.onCancel(GrpcServerExceptionInterceptor.java:63)
	at io.grpc.PartialForwardingServerCallListener.onCancel(PartialForwardingServerCallListener.java:40)
	at io.grpc.ForwardingServerCallListener.onCancel(ForwardingServerCallListener.java:23)
	at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onCancel(ForwardingServerCallListener.java:40)
	at net.devh.boot.grpc.server.security.interceptors.AbstractAuthenticatingServerCallListener.onCancel(AbstractAuthenticatingServerCallListener.java:117)
	at io.grpc.PartialForwardingServerCallListener.onCancel(PartialForwardingServerCallListener.java:40)
	at io.grpc.ForwardingServerCallListener.onCancel(ForwardingServerCallListener.java:23)
[...]

PR #2031, opened 2022-06-10
#1681 Improve error in case of duplicate inserts

uweschaefer:


bresciamattia (approved):
Looks good to me. Thanks @uweschaefer

Issue #1563, opened 2021-09-01
Improve performance during catchup for transformed events


BernhardBln:
**Is your feature request related to a problem? Please describe.**
Currently subscription requests for fact versions that are different than the original, published fact are slow.

Either they need to be transformed using javascript, or they need to be read (one by one) from the `transformationcache` table.

**Describe the solution you'd like**
For a subscription, we fetch `factcast.store.page-size` many records from the database at once. 

Identify from those records the ones that need transformation, and concurrently (and with a single call to the database) look them up from the `transformationcache`, with a javascript transformation as fallback.

The advantage of this solution is that 

- we do not make a single call to the database for each missing fact version, but one to fetch all missing ones in the current batch
- we can look up / transform the missing versions in parallel while the client is processing facts, making good use of the client's processing time

**Describe alternatives you've considered**
There was the suggestion to engrave fact versions (https://github.com/factcast/factcast/issues/1294) but in many scenarios there will not be _the_ perfect choice for a version, and it will require teams to constantly touch old and working code to upgrade event versions.

**Additional context**
See measurements from https://github.com/factcast/factcast/issues/1294, serving transformed facts almost took 66 times longer.

uweschaefer:
reopening

PR #2060, opened 2022-06-27
#1563 bulkread from transformation cache

uweschaefer:


factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -37,97 +43,173 @@ public class PgTransformationCache implements TransformationCache {
...
 
-    cacheKeysBuffer.add(cacheKey);
-    if (cacheKeysBuffer.size() >= maxBufferSize) {
-      CompletableFuture.runAsync(this::flush);
+    List<Fact> facts =
+        keys.stream().map(buffer::get).filter(Objects::nonNull).collect(Collectors.toList());
+
+    if (facts.size() < keys.size()) {
+      NamedParameterJdbcTemplate named = new NamedParameterJdbcTemplate(jdbcTemplate);
+      SqlParameterSource parameters =
```
otbe, line 103: Is there a specific reason why we (try to) fetch all transformed facts from PG instead of only the ones which are missing (from near cache)? 🤔

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -37,97 +43,173 @@ public class PgTransformationCache implements TransformationCache {
...
-            (ps, arg) -> ps.setString(1, arg));
-      } catch (RuntimeException e) {
-        log.warn(
-            "Could not complete batch update last_access on transformation cache. Error: {}",
+        insertBufferedTransformations(copy);
+        insertBufferedAccesses(copy);
+
+      } catch (Exception e) {
+        log.error(
+            "Could not complete batch update of transformations on transformation cache. Error: {}",
```
otbe, line 173: Can we just pass down `e`? 👀

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
uweschaefer: test added

PR #2089, opened 2022-07-20
1817 escalate cancellation of a subscription up into projectorlenses

### LINKED ISSUE #1817: Escalate cancellation of a subscription up into ProjectorLenses
Otherwise we see dirty bulks that are tried to be flushed without an underlying transaction

uweschaefer:
closes #1817


bresciamattia (approved):
LGTM, thanks @uweschaefer

PR #2099, opened 2022-07-25
2098 schema caching fix

### LINKED ISSUE #2098: Profile Fact Validation for performance improvements
When facing mass ingestion, the fact validation is a major factor in publishing throughput.
We'd like to see, if there is something we can do (next to switching to the fastest validator we found working) in order to improve throughput.

uweschaefer:
fixes #2098


otbe (approved):
LGTM

PR #2120, opened 2022-08-05
2119 prevent starving liquibase when replacing function used in trigger

uweschaefer:


bresciamattia (approved):
lgtm, thanks @uweschaefer

Issue #2088, opened 2022-07-20
Suppress "ResultSet closed" Exception caused by canceling the statement


uweschaefer:
When a Statement is cancelled due to a consumer canceling the subscription or disappearing, the Statement is cancelled in order to release resources from the DB asap. When this happens while iterating over the ResultSet, the Thread reading from it would of course be haunted by a "org.postgresql.util.PSQLException: This ResultSet is closed." which is as expected as necessary to break the control flow.
However, it would be nice to suppress this exception (or lower its log level) so that nobody gets alarmed by what is ultimately a "consumer shutting down".

ResultSet::getStatement() might be helpful to help to track it down, event though it must be called before the cancel happens, because it throws for closed RSs.
![image](https://user-images.githubusercontent.com/546815/179952398-05aa3a44-e387-4e48-92b9-3df3aaecf543.png)

PR #2124, opened 2022-08-08
#2088 suppress exception

uweschaefer:


factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -18,22 +18,49 @@
...
-      try {
-        statement.cancel();
-      } catch (SQLException e) {
-        log.debug("Exception while closing statement {}:", statement, e);
+    synchronized (mutex) {
+      if (statement != null) {
+        log.info("Canceling statement " + statement);
+        try {
+          statement.cancel();
+          // also set it to null, so that a thread can see it its statement was canceled
```
bresciamattia, line 37:
```suggestion
```

bresciamattia (approved):
lgtm, thanks @uweschaefer

PR #2169, opened 2022-10-15
#2168: rollback tx in case of statement cancellation

otbe:


factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -33,7 +33,13 @@ public void close() {
       if (statement != null) {
         log.info("Canceling statement " + statement);
         try {
+
           statement.cancel();
+
+          // we have to rollback the tx on the undderlying transaction
```
uweschaefer, line 39: double d

uweschaefer (approved):

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
otbe: Its really that important!!! :D

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -33,7 +33,13 @@ public void close() {
       if (statement != null) {
         log.info("Canceling statement " + statement);
         try {
+
           statement.cancel();
+
+          // we have to roll back the tx on the underlying connection
+          // if we do not end the transaction, statements are cancelled but still "idle in
+          // transaction" and so block further actions like wiping between tests
+          statement.getConnection().rollback();
```
uweschaefer, line 42: wasCanceled should be set to true **before** the rollback is triggered, so that in case of an exception the boolean is set, right?

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
uweschaefer: just trying to help :man_shrugging:

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
otbe: I was thinking of moving it to a finally block 👋

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
otbe: I meant the extra d underlines the importance 😅 adjusted the comment

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
uweschaefer: :laughing:

uweschaefer: even better

otbe: done :)

PR #2269, opened 2023-04-03
2052: Add blacklist fetching from file/ classpath 

bedaka:
Introduces a way to use a local blacklist by pointing to a blacklist file in the filesystem or classpath.

- default remains a blacklist table in postgres
- specify type by setting `factcast.blacklist.type` to either `POSTGRES` or `RESOURCE`
- if a resource based blacklist is used, point to it by setting `factcast.blacklist.location`


factcast-store/src/main/java/org/factcast/store/internal/filter/blacklist/BlacklistConfigurationProperties.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@SuppressWarnings("DefaultAnnotationParam")
+@ConfigurationProperties(prefix = "factcast.blacklist")
+@Slf4j
+@Data
+@Validated
+@Accessors(fluent = false)
+public final class BlacklistConfigurationProperties implements InitializingBean {
```
uweschaefer, line 31: the properties need to be documentend in /factcast-site.

bedaka:
> please add docs

@uweschaefer I added a new section under `Concepts` feel free to adjust it.

uweschaefer (approved):

PR #2270, opened 2023-04-04
2052: remove unused transitive codec dependencies

bedaka:
Against expectations java ignores the lazy class loading when using the method reference operator instead of a lambda which is why building fails for clients using the newest version because of missing dependencies. 
This PR reverts that and removes the now not longer required test dependencies.


uweschaefer (approved):

uweschaefer:
@bedaka feel free to merge once pipeline is green

Issue #2280, opened 2023-05-09
Make factcast-hugo create a version that works with MACs


uweschaefer:

uweschaefer:
closed by #2281

PR #2281, opened 2023-05-09
#2280: Replace factcast-hugo by public variant

uweschaefer:
Still it needs to be built on a linux machine for it to work on macs. home-built factcast-hugo on mac do not work


uweschaefer:
closes #2280

uweschaefer:
LGTM

PR #2317, opened 2023-06-20
#2316-tx-manager: fixed handling of tx manager in redis transactional lens

### LINKED ISSUE #2316: RedisTransactionalLens stores wrong tx manager instance
We ran into a strange exception in event handlers of transactional subscribed redis projections, when running integration tests (using a test extension that subscribes subscriptions on demand, per test):

> Unable to execute operation. Transaction is in finished state!

The error comes when trying to obtain a redis map or set from the given RTransaction.

The reason is apparently that the RedisTransactionalLens keeps a copy of the txManager when created.

In our test setup (not sure if its the same in general) the lenses for **all** projections are created on the same thread, the one that also discovers event handler methods.

And because of this they all share **the same** transaction manager, and when the subscription happens after (on different threads per projection with dedicated tx managers per projection), the lenses will still use the shared transaction manager from above.

Worse, the abstract redis projections use the dedicated txManager for the subscription thread, so the event handlers get the shared tx manager injected (and change their data there), but the update to the state is happening on the individual tx managers.

Latest.

**To Reproduce**

Create two redis subscribed projections. You most likely run into that error (but it depends on concurrency, if you don't see it try multiple times)

**Expected behavior**
We don't run into `Unable to execute operation. Transaction is in finished state!` when getting redis maps or sets from the RTransaction.

BernhardBln:
Fixes #2316


BernhardBln:
Tests are all green locally, by the way.

PR #2320, opened 2023-06-21
#2316-state-bucket-protected: Make stateBucket protected and hence av…

BernhardBln:
…ailable to subclasses


BernhardBln:
We'll take a better approach where we keep the bucket encapsulated.

PR #2323, opened 2023-06-26
rename ProjectionMetaData::serial -> revision

### LINKED ISSUE #2322: rename ProjectionMetaData.serial
As serial is a heavy used term within factcast, it's probably a good idea to express that this is about something diffferent.
Also we avoided to use serialVersionUID here, as - while having smilar semantics, it might differ from what you want to express in the annotation.

we settled for projectionVersion

uweschaefer:
fixes #2322


uweschaefer:
added compatibility code (tmp until 0.8) to not break clients

uweschaefer:
superseeded by #2541

PR #2324, opened 2023-07-10
Adding helper class to translate from EventObject to a Fact

mamup11:


factcast-test/src/main/java/org/factcast/test/FactFromEvent.java
```diff
@@ -0,0 +1,57 @@
...
+
+    if (specs == null) {
+      throw new IllegalStateException("invalid event object");
+    } else {
+      return Fact.builder()
+          .type(specs.type())
+          .ns(specs.ns())
+          .version(specs.version())
+          .id(id)
+          .meta("_ser", String.valueOf(serial));
```
uweschaefer, line 54: would be nice if we returned a custom builder that adds a serial(int) call?

factcast-test/src/main/java/org/factcast/test/FactFromEvent.java
```diff
@@ -0,0 +1,57 @@
...
+  public static Fact.Builder factFromEvent(@NotNull EventObject event) {
+    return factFromEvent(event, 0);
+  }
+
+  public static Fact.Builder factFromEvent(@NotNull EventObject event, long serial) {
+    return factFromEvent(event, serial, UUID.randomUUID());
+  }
+
+  public static Fact.Builder factFromEvent(@NotNull EventObject event, long serial, UUID id) {
+    Annotation[] annotations = event.getClass().getAnnotations();
```
uweschaefer, line 36: why not just getAnnotation(Specification.class) ?

factcast-test/src/main/java/org/factcast/test/FactFromEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      return Fact.builder()
+          .type(specs.type())
+          .ns(specs.ns())
+          .version(specs.version())
+          .id(id)
+          .meta("_ser", String.valueOf(serial));
+    }
+  }
+}
```
uweschaefer, line 57: i kind of miss the payload here :D

uweschaefer (changes requested):

uweschaefer:
superseeded by #2329

PR #2334, opened 2023-08-22
#2333: Read-only configuration

### LINKED ISSUE #2333: Allow read-only deployment of factcast server
It would be nice if we were able to configure the factcast server in such a way, that it operates in a read only mode. 
Could be quite useful if you want to deploy a dedicated eventstore that uses a read replica for reporting reasons for example.

(As of now) That means:
- no event publishing
- no liquibase
- scheduled actions:
  - tail index manager should be off
- in mem snapshot cache
- persistent schema-reg/transformationstore should be usable but read-only
- persistent transformationcache should be usable but read-only
- throwing tokenstore

otbe:
closes #2333


uweschaefer (approved):

bedaka (commented):
Only found some typos in comments.

PR #2341, opened 2023-08-31
#2340: Possbile OOM/buffer overflow in BufferingFactInterceptor

### LINKED ISSUE #2340: Possible OOM/buffer overflow in BufferingFactInterceptor in heavy load sitautions
Especially if you have some facts that need transformations and afterwards many that dont need transformation.
When the BufferingFactInterceptor entered buffering mode and afterwards only accepts facts that don't need transformation no one calls `flush` when the max size is reached. See [this](https://github.com/factcast/factcast/blob/5b12b3c09ee2aec6b802a53ad715fa6435fb3c06/factcast-store/src/main/java/org/factcast/store/internal/catchup/BufferingFactInterceptor.java#L88C18-L88C41).

Observed via customer report&otel tracing.

otbe:
closes #2340


otbe:
mem consumption before ~unlimited afterwards ~200mb while processing 20m facts 😎

uweschaefer (approved):
lgtm

Issue #2335, opened 2023-08-22
Update java version for sonar job


otbe:
Sonar deprecates j11 soon'ish and we should update the java version used for running sonar in our github actions.

PR #2482, opened 2023-10-09
#2335: JDK update 11->17 in maven workflow

leflamm:


leflamm:
⚠️ Please do not merge yet ⚠️

### consistently updated JDK versions in `workflows/maven.yaml` file from v11 to v17
* all tests are green
* famous: `Warning:  The version of Java (11.0.20.1) you have used to run this analysis is deprecated and we will stop accepting it soon. Please update to at least Java 17.` is gone
* after diffing the build output between a j17 and a previous j11 run I noticed 
  * in `factcast-schema-registry-cli` we now have a few previously unseen `Warning: [WARNING] warning: Supported source version 'RELEASE_16' from annotation processor 'org.jetbrains.kotlin.kapt3.base.ProcessorWrapper' less than -source '17'` which could indicate that the involved plugin does not care for any previously set source/target=11 value 
  * an occasional `[WARNING] system modules path not set in conjunction with -source 11` when (test) compiling, which AFAUI appears if you use a JDK and build with different source/target compatibility value

.github/workflows/maven.yaml
```diff
@@ -46,12 +46,24 @@ jobs:
       name: 'Test - Unit'
       run: './mvnw -B test'
     - id: 'step-6'
+      name: 'JDK 17 - only for Sonar'
```
otbe, line 49: This file is generated via the kts file in `.github/kts/maven.main.kts. I would not change it directly :D

Switch to j17 is also done on the 0.7/sb3 branch and should be merged soon, so I would recommend to not merge this PR sadly :(

leflamm: all good, just playing around, no not merge! :-)

leflamm:
Some converted itself into a play-around branch... its current content is of no value!

.github/workflows/maven.yaml
leflamm: Somehow converted itself into a play-around branch... its current content is of no value!

PR #2483, opened 2023-10-09
Adding enumeration capabilities to schema registry interface and impl…

mamup11:
## Description
This PR adds a method to enumerate the namespaces and the types of events to the schema registry contract, to be able to fetch this information from the registry without having to go to the store when the registry is present


uweschaefer:
thx @mamup11

PR #2487, opened 2023-10-10
2484: remove unnecessary duration assertion

leflamm:


factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/SubscriptionReconnectionITest.java
```diff
@@ -114,12 +108,7 @@ void subscribeWithReconnect() {
           count.incrementAndGet();
         });
 
-    long rtWithReconnect = sw.stop().elapsed(TimeUnit.MILLISECONDS);
-
-    assertThat(rtWithReconnect).isGreaterThan(rt);
```
leflamm, line 119: going bs=1 one cannot assume that one 100ms sleep will definitely make the catchup for the catchup-with-reconnect slower in every case, also it doesn't seem important to me

bedaka (approved):

PR #2521, opened 2023-10-14
UI playground

otbe:


uweschaefer:
codefactor warnings can be ignored. they do not even show up on codefactor.io - if this persists, we will just remove it from the project

PR #2540, opened 2023-10-25
Add dependency declaring interface in order to ensure init/destroy order in Spring

### LINKED ISSUE #2381: "Redisson already shut down" error messages on app shutdown
**Describe the bug**
We see a lot of error messages in the log when an application is gracefully shut down, that redisson is already shut down.

Most likely because factus has an implicit dependency on redisson: The bean container is not aware of that dependency, but factus keeps a couple of resources (that are not beans) that will be closed upon destruction of factcast, and some of those resources require redisson to stay active until they got closed.

0.6.9

**To Reproduce**
Difficult, as it is a bit random, but happens usually on AWS / ECS when redeploying

**Expected behavior**
No error messages any more, as redisson is only shut down after all resources that depend on it got shut down before.

uweschaefer:
closes #2381 
superseeds #2383 

This way, people can either implement, or use an indirection (see RedissonAutoConfiguration as an example)  to declare dependencies to Factus the Spring way, without necessarily changing Factcast code.

### DIFF
--- a/factcast-spring-boot-autoconfigure/src/main/java/org/factcast/spring/boot/autoconfigure/redis/RedissonAutoConfiguration.java
+++ b/factcast-spring-boot-autoconfigure/src/main/java/org/factcast/spring/boot/autoconfigure/redis/RedissonAutoConfiguration.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.factcast.spring.boot.autoconfigure.core;
+package org.factcast.spring.boot.autoconfigure.redis;
 
 import org.factcast.factus.utils.FactusDependency;
 import org.redisson.api.RedissonClient;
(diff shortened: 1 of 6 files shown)


BernhardBln (approved):
Looks good

factcast-spring-boot-autoconfigure/src/main/java/org/factcast/spring/boot/autoconfigure/redis/RedissonAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import org.redisson.api.RedissonClient;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.context.annotation.Bean;
+
+@AutoConfiguration
+@ConditionalOnClass({RedissonClient.class})
+public class RedissonAutoConfiguration {
+
+  @Bean
```
otbe, line 28: Technically `ConditionalOnClass` does not mean that a bean was created right?
I wonder if we should add a `ConditionalOnBean` to the bean providing method

BernhardBln: True, good point.

PR #2541, opened 2023-10-25
#2322_revision: renamed to revision while keeping compatible

uweschaefer:


uweschaefer:
added compatibility code (tmp until 0.8) to not break clients

uweschaefer:
closes #2322

otbe (approved):
LGTM :)

PR #2544, opened 2023-10-25
WIP: Solve coverage disagreement between jacoco and sonar

uweschaefer:
Rest assured, merging into ui_playground is correct

image


uweschaefer:
Turns out that the reason for excluded class to still show up (as Files) in the sonarcloud-coverage-list-of-shame was caused by inner classes (also anonymous ones and even lambdas) being not explicitely excluded from coverage recording.

PR #2692, opened 2024-01-23
#2691: add support for separate credentials

### LINKED ISSUE #2691: Provide a way to specify user&password in two different properties
This would enable us to provide user&pw via something like [awsspring](https://docs.awspring.io/spring-cloud-aws/docs/3.1.0/reference/html/index.html#spring-cloud-aws-secrets-manager) and also via mechanisms like [ECS Secrets](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar.html). Additonally, we can use the same source on the server side where we have to specify `user: password` via properties or programmatically.

One could easily provide a backwards compatible way by introducing `factcast.grpc.client.user` and `factcast.grpc.client.password` and checking both places.

otbe:
closes #2691


uweschaefer:
otherwise lgtm

uweschaefer (approved):

Issue #2578, opened 2023-11-15
Handshake on GRPC failover


uweschaefer:
**Describe the bug**
When a client fails over to another server, there currently is no new handshake, which is a problem as the server config can be different from the original server (protocol version, compression etc), so that the handshake has to happen for a fresh start of communication to the new server.

bedaka:
This can be tested by deploying the server and verifying if there are handshake logs from the client.

Place for change is within the reconnect logic. 

somewhere in:
ResilientGrpcSubscription

PR #2693, opened 2024-01-25
#2578 handshake before reconnect

bedaka:


bedaka:
was moved over to: https://github.com/factcast/factcast/pull/2737

Issue #2832, opened 2024-04-08
Flaky: ExceptionHandlingV4.testSubscriptionFC_transformationErrors_follow


leflamm:
```
[ERROR] Tests run: 10, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 53.10 s <<< FAILURE! -- in org.factcast.itests.exception.handling.ExceptionHandlingV4
[ERROR] org.factcast.itests.exception.handling.ExceptionHandlingV4.testSubscriptionFC_transformationErrors_follow -- Time elapsed: 2.425 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting value to be true but was false
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at org.factcast.itests.exception.handling.ExceptionHandlingV4.testSubscriptionFC_transformationErrors_follow(ExceptionHandlingV4.java:221)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)

INFO  org.factcast.test.AbstractFactCastIntegrationTest - STDOUT: INFO  org.factcast.store.internal.query.CurrentStatementHolder - Canceling statement org.apache.tomcat.jdbc.pool.StatementFacade$StatementProxy[Proxy=978176964; Query=SELECT ser, header, payload, header->>'id' AS id, header->>'aggIds' AS aggIds, header->>'ns' AS ns, header->>'type' AS type, header->>'version' AS version FROM fact WHERE ( (1=1 AND header @> ?::jsonb AND header @> ?::jsonb) ) AND ser>? ORDER BY ser ASC; Delegate=null]
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   ExceptionHandlingV4.testSubscriptionFC_transformationErrors_follow:221 
Expecting value to be true but was false
[INFO] 
[...]

PR #2835, opened 2024-04-08
#2832: relax errorLatch assertion timeout

leflamm:
Relax timeout value ~and checking behavior~ after noticing it was too strict for some machines (mine), see https://github.com/factcast/factcast/issues/2832


leflamm:
Realized that CountDownLatch.await(...) already behaves as desired
image

uweschaefer:
Extracted constant. LGTM, thx @leflamm

uweschaefer (approved):

Issue #2868, opened 2024-05-03
Re-initialization causes duplication of gRPC headers


bresciamattia:
**Describe the bug**
Upon re-initialization of the GrpcFactStore (e.g. during resilient subscription reconnect due to fc server being deployed), the gRPC headers of the handshake are duplicated.
This could result in the client not being able to connect again, due to header size exceeded.

**FactCast Server/Client Version**
0.7.5 to 0.7.6

**To Reproduce**
Trigger onError on a subscribed projection to resubscribe, multiple times.

**Expected behavior**
The gRPC message headers should not be duplicated.

PR #2872, opened 2024-05-07
#2868: Re-initialization causes duplication of gRPC headers

bresciamattia:


uweschaefer (approved):
Mille Grazie @bresciamattia

Issue #2723, opened 2024-02-08
Replace jpountz lz4 by maintained lz4 version (commons-compress?)


uweschaefer:

uweschaefer:
add regression test

uweschaefer:
closed by #2916

PR #2916, opened 2024-05-21
#2723: added commons lz4 codecs

uweschaefer:


uweschaefer:
i guess we need a property for the client to express priorities, now that he cannot do this any more by selecting different dependencies

factcast-grpc-api/src/main/java/org/factcast/grpc/api/CompressionCodecs.java
```diff
@@ -32,9 +32,11 @@ public class CompressionCodecs {
 
   private final String orderedListOfAvailableCodecsAsString;
 
+  private static final String[] AVAIL_CODECS = {"lz4c", "snappyc", "lz4", "snappy", "gzip"};
```
leflamm, line 35: add explanation for `FOO` vs. `FOOc` ?

leflamm (commented):
LGTM! Tiniest amount of suggestions. Cheers!

For me merging into master would make sense.

Maybe `CompressionCodecs` could use a UT.

Issue #2931, opened 2024-05-31
Make sure if using separate username/password, that both are set on the client side


uweschaefer:
**Describe the bug**
Currently clients can get away with setting only one of password/username, which the later causes an exception in the server.

This needs to be caught early.

PR #2939, opened 2024-06-03
#2931 fail fast on missing credential parts

leflamm:
Missing (and, as discussed with @uweschaefer, also empty) credential parts should cause fast failure on the client side.


uweschaefer (approved):
thx @leflamm

PR #2947, opened 2024-06-04
#2882: PgStoreTelemetry

### LINKED ISSUE #2882: Use the guava's EventBus to expose telemetry data from factstore
It would be nice to be able to observe details within the factstore, in order to analyze and log.

Being able to listen to the event bus from the server implementation for this telemetry data.
E.G.: receive events from the bus whenever a subscription query starts/ends, with each subscription request.

This should be documented, to allow users to subscribe to those internal events.

bresciamattia:
closes #2882


factcast-store/src/main/java/org/factcast/store/internal/telemetry/PgStoreTelemetry.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private final EventBus eventBus;
+
+  public PgStoreTelemetry() {
+    this(
+        new AsyncEventBus(
+            PgStoreTelemetry.class.getSimpleName(),
+            // needs to be a singleThread executor to ensure that the order of events is
+            // respected
+            Executors.newSingleThreadExecutor()));
```
otbe, line 36: I wonder if it would be valuable to instrument the ES via micrometer. We could see some metrics in our dashboards about how many signals are dispatched/waiting times. Same could be done for the old eventbus. Wdyt?

bresciamattia: Yes that would definitely make sense. It would be a lot easier to troubleshoot possible bottlenecks.

otbe (approved):
LGTM

uweschaefer:
Thx @bresciamattia

Issue #2944, opened 2024-06-04
Flaky test: PgFactStoreIntegrationTest.testSerialAndTimestampWereAugmented


otbe:
More often than not this test fails on my machine (Mac)

```
[ERROR]   PgFactStoreIntegrationTest.testSerialAndTimestampWereAugmented:155 
Expecting actual:
  1717504295436L
to be greater than or equal to:
  1717504295437L
```

Issue #2949, opened 2024-06-04
Flaky Test: ClientWithSeparateCredentialsTest.failsToCatchup 


uweschaefer:
**Describe the bug**

https://github.com/factcast/factcast/actions/runs/9374545904/job/25810816870?pr=2915

or

https://github.com/factcast/factcast/actions/runs/9560351921/job/26352321252?pr=2964

leflamm:
some progress... at least reproducible now:

In `ClientWithSeparateCredentialsTest` comment-out all test methods other than `failsToCatchup` and run via 

```bash
~/git/hub/factcast/factcast-itests/factcast-itests-security$ mvn clean -DskipUnitTests verify -Dit.test=ClientWithSeparateCredentialsTest 
```
This fails every time (on my machine :-/)

Weird, though:
* does not fail in Idea
* does not fail if other tests are `@Ignore`d

leflamm:
Managed to reproduce the error locally:
- it is a race condition depending on whether the `ResilientGrpcSubscription#awaitCatchup` happens before or after `ResilientGrpcSubscription#close`, also see https://github.com/factcast/factcast/pull/2968
- in the 'after' case a generic `SubscriptionClosedException` hides the actual error-causing Throwable, here: `StatusRuntimeException` expressing `PERMISSION_DENIED`
- provided solution: remember the error-causing Throwable and re-throw it in this situation, wipe that Throwable in case of a successful reconnect

PR #2964, opened 2024-06-17
#2944: respect os-db time drift

leflamm:
Trying to take a potential time drift into account


leflamm:
some IT hiccups because of #2949 :-/

PR #3006, opened 2024-07-02
#3005: remove stale retry methods

### LINKED ISSUE #3005: Remove deprecated methods from ReadFactCast API
**Describe the bug**

retry methods deprecated since 0.5.6 can be removed from `ReadFactCast`.

uweschaefer:
closes #3005


factcast-core/src/main/java/org/factcast/core/DefaultFactCast.java
```diff
@@ -78,11 +78,11 @@ public Set<String> enumerateTypes(@NonNull String ns) {
   @Override
   @NonNull
   @SuppressWarnings("deprecated")
```
otbe, line 80: Can go away I think? Nothing deprecated is used

otbe (approved):
LGTM :)

PR #3047, opened 2024-07-17
Issue2330 add meta parameter injection

### LINKED ISSUE #2330: Add @Metadata("foo") String bar parameter injection into handler methods
Right now, the only way to get hold of custom metadata objects is to inject the Fact as a parameter to handler methods.
As the meta header Attribute is a Map<String,String> we could easily allow to access selected meta data fields by allowing to inject them into handlers, like 
```java
void apply(MyEvent e, @Metadata("myMetaField") String field1, @Metadata("myOtherMetaField") String field2){...}
```

Of course we could offer implicit conversion, for instance to Long or long, but what if there is a conversion error? There is a trade-off between convenience and robustness here. Also we'd need to be able to inject Optional<Long> or OptionalLong as the values might be null.

uweschaefer:
closes #2330


factcast-factus/src/main/java/org/factcast/factus/projection/parameter/DefaultHandlerParameterContributor.java
```diff
@@ -24,35 +28,134 @@
...
+      return new FactHeaderProvider();
     }
 
     if (UUID.class == type) {
-      return (f, p) -> f.id();
+      return new FactIdProvider();
     }
 
-    if (FactStreamPosition.class == type) return (f, p) -> FactStreamPosition.from(f);
+    // TODO is this documented already?
```
bedaka, line 61: follow up ticket required?

otbe (approved):

bresciamattia (approved):
LGTM!

PR #3068, opened 2024-07-29
metaExists / DoesNotExist

### LINKED ISSUE #2814: Allow Fact Spec to filter for "meta key exists" instead of key/value pair only
We have a use-case where an fact `E` has multiple variants:
 - one variant does not set meta-data at all
 - one variant sets meta-data `key` to `value1`
 - one variant sets meta-data `key` to `value2`
 - ...

A way to build a fact stream that returns all `E`s with meta-data `key` set at all, no matter the value, so some kind of `exists(key)` functionality for the meta-data

While we're at it, we could also do the opposite like 
* metaDoesNotExist("key") 
* metaExists("key") 

Alternatives I have considered, but dislike:
 - adding a `hasKey: true` key-value pair to every variant setting `key` to any value
 - Doing the meta-data filtering on the client-side 

**Additional context**
@uweschaefer this is what we talked about yesterday 😁

uweschaefer:
closes #2814


uweschaefer:
note that this shifts the PG default to 15 (12 would have been needed)

otbe:
We might want to update https://docs.factcast.org/setup/prerequisites/.
Is there another section where we promise pg11 (and not 9.4 like above :D)?

otbe (approved):
Looks as awesome as it can be :)

Issue #3176, opened 2024-10-07
Fix Mockito self-attaching


leflamm:
**Is your feature request related to a problem? Please describe.**
Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. 

```
WARNING: A Java agent has been loaded dynamically (/home/cf/.m2/repository/net/bytebuddy/byte-buddy-agent/1.15.3/byte-buddy-agent-1.15.3.jar)
WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
WARNING: If a serviceability tool is not in use, please run with -Djdk.instrument.traceUsage for more information
WARNING: Dynamic loading of agents will be disallowed by default in a future release
```

**Describe the solution you'd like**
Please add Mockito as an agent to your build as described in Mockito's documentation: [Mockito - mockito-core 5.14.1 javadoc](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3)

**Additional context**
* Note that e.g. [`CommandServiceImplTest`](https://github.com/factcast/factcast/blob/master/factcast-schema-registry-cli/src/test/kotlin/org/factcast/schema/registry/cli/commands/CommandServiceImplTest.kt) uses MockK which causes the same warning but will not be fixed with suggested solution
* open MockK issue: https://github.com/mockk/mockk/issues/1171 
* a rather butchery option here would be to replace the entire MockK usage with Mockito, which I do not want to do

PR #3177, opened 2024-10-07
#3176: add Mockito lib as java agent

leflamm:
Adding Mockito lib as Java agent because starting from Java 21, the [JDK restricts the ability of libraries to attach a Java agent to their own JVM](https://openjdk.org/jeps/451). As a result, the inline-mock-maker might not be able to function without an explicit setup to enable instrumentation, and the JVM will always display a warning.

Note that e.g. [CommandServiceImplTest](https://github.com/factcast/factcast/blob/master/factcast-schema-registry-cli/src/test/kotlin/org/factcast/schema/registry/cli/commands/CommandServiceImplTest.kt) uses MockK which causes the same warning but will not be fixed with the suggested solution.

MockK still has an open issue regarding this, see https://github.com/mockk/mockk/issues/1171

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,6 +500,17 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>properties</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>${maven-surefire-plugin.version}</version>
         <configuration>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -498,6 +498,17 @@
             unified report if you do mvn verify -->
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>properties</goal>
```
leflamm, line 507: fyi: 

> Goal that sets a property pointing to the artifact file for each project dependency. For each dependency (direct and transitive) a project property will be set which follows the groupId:artifactId:type:[classifier] form and contains the path to the resolved artifact

https://maven.apache.org/plugins/maven-dependency-plugin/properties-mojo.html

pom.xml
```diff
@@ -508,7 +519,8 @@
           </includes>
           <excludedGroups>integration,uitest</excludedGroups>
           <!--suppress UnresolvedMavenProperty -->
-          <argLine>${surefire.jacoco.args}</argLine>
+          <argLine>${surefire.jacoco.args}
+            -javaagent:${org.mockito:mockito-core:jar}</argLine>
```
leflamm, line 523: fyi: making use of that dependency property

uweschaefer (approved):

PR #3244, opened 2024-10-31
Update AbstractFileBasedRegistryFileFetcher.java

Cgarg547:
I have made a change in the abstract class which will allow fetching the files from different sources flexibly. Also, the code will throw the exceptions properly in order to provide error messages.


uweschaefer:
Hi @Cgarg547 and welcome to the project.

We value new contributors very much.

That being said, javadoc/exception improvements aside, i don't exactly get you change here.
How does your refactoring make the fetching more flexible? Did you mean to make the buildXYZ methods protected so that they can be overridden?

Maybe you could provide an issue that gives a bit more context for this change?

Thanks

PR #3256, opened 2024-11-04
Issue3255 retry on cancel

bedaka:


factcast-client-grpc/src/main/java/org/factcast/client/grpc/ClientExceptionHelper.java
```diff
@@ -75,9 +76,20 @@ public static boolean isRetryable(@NonNull Throwable exception) {
...
+      return RETRYABLE_STATUS.contains(s) || isNotCausedByClient(runtimeException);
     }
     return false;
   }
+
+  /**
+   * Check for message that indicates an issue caused by a proxy (e.g. LoadBalancer) and therefore
+   * warrants a retry.
+   */
+  private boolean isNotCausedByClient(StatusRuntimeException e) {
```
uweschaefer, line 90: Might be easier to read if we'd rename it to not include a negation.

uweschaefer:
```suggestion
  private boolean isCausedByInfrastructure(StatusRuntimeException e) {
```

uweschaefer (approved):

PR #3267, opened 2024-11-08
Add support for multiple aggIds in FactSpecs

### LINKED ISSUE #1093: Allow multiple aggIds in FactSpec
Right now we can add as many aggIds as we want when publishing facts but on the read side we only can specifiy one aggId in a `FactSpec`. 

I think it would be nice to be able to specify multiple aggIds in a `FactSpec`

:)

bresciamattia:
Closes #1093

### DIFF
--- a/factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
+++ b/factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
@@ -50,9 +50,6 @@ public class FactSpec {
 
   @JsonProperty private int version = 0; // 0 means I don't care
 
-  // hide aggId getter / setter as it's deprecated to use and replaced by aggId_s_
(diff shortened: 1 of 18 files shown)


factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java
```diff
@@ -48,6 +52,22 @@ public class FactSpec {
 
   @JsonProperty FilterScript filterScript = null;
 
+  public Set<UUID> aggIds() {
```
uweschaefer, line 55: unexpected behavior when used to modify from the outside.
We should wrap into unmodifiable

uweschaefer (approved):

uweschaefer:
thx @bresciamattia

PR #3291, opened 2024-11-21
#3289: FastForward information might be stale but is still applied client-side

### LINKED ISSUE #3289: FastForward information might be stale
Depending on the timing, the fastForward that is send to clients might be stale. Imagine a situation like this:

E1 -> ser=1
E2 -> ser=2
-> FastForwardTargetRefresher refreshes and now has targetSer=2
E3 -> ser=3
E4 -> ser=4

If you build a aggregate that reads these events it will receive a fastForward at the end of handling E4 (which is new due to server pipelines, before it got it basically right in the beginning thats why this issue did not happen in <0.8).
The code currently then sets the `positionOfLastFactApplied` to the value of 2 in `org.factcast.factus.FactusImpl#catchupProjection`. That means if you fetch this aggregate again event E3 and E4 are applied again!

This happens more often in a test setup where the serials are not reset between test runs but can be also an issue for production use cases (we refresh the target ser only every 5 minutes).

Even worse, this can also affect subscribed projections.

I think for all kinds of projections we can fix it client-side by checking if the serial in ffwd is bigger than the one from the last handled event but we might also have to change the server implementation to be on the safe side.

otbe:
- ignore stale ffwd in factus
- only send ffwd if its really a forward

closes #3289

### DIFF
--- a/factcast-core/src/main/java/org/factcast/core/FactStreamPosition.java
+++ b/factcast-core/src/main/java/org/factcast/core/FactStreamPosition.java
@@ -43,4 +43,15 @@ public class FactStreamPosition {
     if (uuid == null) return null;
     else return FactStreamPosition.of(uuid, -1L);
   }
+
+  /**
+   * Checks if fsp is after the current one. Its considered being after if the `factStreamPosition`
+   * is null (no FSP yet) or when the serial of the current FSP is greater than the one given.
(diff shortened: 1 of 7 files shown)


uweschaefer:
:scream: thx @otbe

Issue #2796, opened 2024-03-13
[Server-UI] Support end-serial


donaldpipowitch:
**Is your feature request related to a problem? Please describe.**
I'd like to create shareable links to historic analysis. Therefor I'd like to include an `end-serial` param (similar to the existing `starting-serial`). With that I could create a query which goes for example from the first of January to the first of February.

**Describe the solution you'd like**
An optional "Ending Serial" input next to the "Starting Serial" would be nice.

**Describe alternatives you've considered**
Even better alternatives would be using dates directly, so that I could share the query params even across different stages where I want to run the same query, but where different serials are used (e.g. `{ from: '2024-01-01T00:00:00.000Z', to: '2024-02-01T00:00:00.000Z' }`). (And the top notch solution would support relative times for queries like "last 24h".)

**Additional context**
Thanks! 😄 🦦

PR #3310, opened 2024-12-02
2796: Adding new filter by ending serial

mamup11:


factcast-server-ui/src/main/java/org/factcast/server/ui/adapter/ListObserver.java
```diff
@@ -27,16 +28,23 @@
...
+  public ListObserver(Long untilSerial, int limit, int offset) {
+    this.untilSerial = untilSerial;
+    this.limit = limit;
+    this.offset = offset;
+  }
+
   @Override
   public void onNext(@NonNull Fact element) {
-    if (isComplete()) {
+    if (isComplete() || hasReachedTheLimitSerial(element)) {
```
uweschaefer, line 47: minor: looks like this check should be moved to isComplete

uweschaefer:
closes #2796

PR #3334, opened 2024-12-11
Issue3131 factcast UI event export

bedaka:


uweschaefer:
Screenshot from 2025-01-13 11-00-21

leads to 

Screenshot from 2025-01-13 11-01-16

which is not wrong, but a little thin on user feedback :)

uweschaefer:
Screenshot from 2025-01-13 11-03-29

shouldn't this be prevented? Why are there even Facts found?

uweschaefer:
Is the end serial supposedly exclusive?
Screenshot from 2025-01-13 11-14-33

I was expecting this to return fact(ser=4), but it does not

uweschaefer:
We should be explicit about which serial is inclusive, and which one is not:
Screenshot from 2025-01-13 11-18-38

uweschaefer:
> We should be explicit about which serial is inclusive, and which one is not

changed the semantics of the "start ser" field to be inclusive

factcast-docker/pom.xml
```diff
@@ -105,7 +105,6 @@
                 <buildx>
                   <platforms>
                     <platform>linux/amd64</platform>
-                    <platform>linux/arm64</platform>
```
otbe, line 108: :(

leflamm: just a mistake

factcast-server-ui/src/main/java/org/factcast/server/ui/report/ReportQueryPage.java
```diff
@@ -0,0 +1,304 @@
...
+        displayWarning(
+            "No data was found for this query and therefore report creation is skipped.");
+      }
+    } catch (ValidationException e) {
+      Notifications.warn(e.getMessage());
+    } catch (Exception e) {
+      Notifications.error(e.getMessage());
+    }
+    // Not re-enabling the queryReportBtn to not incentivise users to generate it
+    // multiple times.
```
bresciamattia, line 274: should we consider re-enabling it at least when the report is empty or, even better, when the filter is updated somehow? Otherwise one would need to reset the filter each time, or know that changing the report name actually re-enables the button (which seems more like a workaround)

factcast-server-ui-s3-reportstore/src/main/java/org/factcast/server/ui/s3reportstore/S3ReportStoreConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import software.amazon.awssdk.services.s3.presigner.S3Presigner;
+import software.amazon.awssdk.transfer.s3.S3TransferManager;
+
+@Configuration
+public class S3ReportStoreConfiguration {
+
+  @Bean
+  @ConditionalOnMissingBean
+  public S3AsyncClient s3Client() {
+    return S3AsyncClient.create();
```
bresciamattia, line 32: if I git it correctly, the whole purpose of using the S3TransferManager and having the aws-crt dependency, is to use its automatic parallel transfer feature (multipart upload/ download). For that to work out of the box, should we use `S3AsyncClient.crtCreate()`, `S3AsyncClient. crtBuilder().build()` or without aws-crt `S3AsyncClient.builder().multipartEnabled(true).build()`?

See https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/transfer-manager.html

bresciamattia (approved):
LGTM @bedaka! Awesome job

PR #3396, opened 2025-01-13
#3361: do not throw for parallel update calls of tx redis projections on same instance

### LINKED ISSUE #3361: Factus throws an error when projection being updated simultaneously by two threads
When two threads that use same projection instance try to update that projection, one of them will succeed and the other will throw an exception: `org.factcast.factus.projection.tx.TransactionAlreadyRunningException: Transaction already running`

I was able to reproduce it with a simple test like this (I removed project specific data):

`@Slf4j
class Test extends AbstractIntegrationTest {

  @Inject
  Factus factus;
@Inject RedissonClient redisson;
  Projection testProj;

  @org.junit.jupiter.api.Test
  @SneakyThrows
  void testProjection() {
    testProj = new Projection(redisson);
    int numberOfThreads = 2; // Number of threads to create
    for (int i = 0; i < numberOfThreads; i++) {
      Thread thread = new Thread(new MyTask());
      getEvents();
      thread.start(); // Start the thread
    }

    Thread.sleep(100000);
  }

  class MyTask implements Runnable {
    @Override
    public void run() {
      System.out.println("Thread " + Thread.currentThread().getName() + " is running the task.");
        factus.update(testProj);
      System.out.println("Thread " + Thread.currentThread().getName() + " completed the task.");
    }
  }

  private void getEvents() {
	//this method should publish 10 events
  }

  @ProjectionMetaData(revision = 1L)
  class Projection extends AbstractRedisTxManagedProjection {
   
    public Projection(@NonNull RedissonClient redisson) {
      super(redisson);

    }

    @Handler
    void apply(EventPojo e)  {
      log.info("*event*");
    }
  }
}`

In our project this happened when dealing with scheduled tasks, that share one repository instance, that uses one projection instance. Instead of throwing an error, I would rather have factus skip the other update process, which is not necessary anyway.

otbe:
We already rely on the redis lock to make sure we dont run multiple updates in parallel (across several instances). When we remove the assertion of the running tx the same applies then to multiple update calls for the same jvm/spring context/projection instance.

closes #3361

### DIFF
--- a/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
+++ b/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
@@ -771,7 +771,7 @@ class FactusClientTest extends AbstractFactCastIntegrationTest {
   }
 
   @Test
-  void testParallelUpdateCallsForRedisTxManaged() {
+  void testParallelUpdateCallsForRedisTxProjections() {
     final var blockingRedisManagedUserNames = new BlockingRedisTxManagedUserNames(redissonClient);
     factus.publish(new UserCreated("John"));
 
--- a/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
+++ b/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
@@ -29,6 +29,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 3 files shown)


factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
```diff
@@ -768,4 +769,45 @@ void testOverriddenNsSubscription() throws Exception {
...
+              sleep(250);
+              return executeUpdate.get();
+            });
+
+    assertThat(u1.thenCombine(u2, (b1, b2) -> b1 && b2))
+        .succeedsWithin(Duration.ofSeconds(1))
+        .isEqualTo(true);
+  }
+
+  private static void sleep(long ms) {
```
uweschaefer, line 806: sneaky

uweschaefer (approved):

Issue #2841, opened 2024-04-10
[Factus] A WriterToken that gets invalid during normal operations leads to a dead subscription


otbe:
In `org.factcast.factus.FactusImpl#doSubscribe` we check the validity of the WriterToken on each fact that arrives. However as soon as the WriterToken is not valid anymore we throw an IAE which basically ends the subscription. A WriterToken which expires is perfectly fine and we should just try to renew it. 

Theres another aspect to this. The quite common case of handling a network related issue (failover or real outtage) is currently done by `ResilientGrpcSubscription` however it keeps the WriterToken alive the whole time. One could argue it should release the lock and give other instances a chance to take over. However the amount of retries is limited and even if the network does not come back at some point another instance takes over.

Still we should think about a proper solution :)

mamup11:
After digging into this a bit and chatting with others, a couple of things came up:

1. Re-acquiring the WriterToken is tricky.
If we try to do it inside onNext, we’d need to compare the FactStreamPosition to make sure no other instance made progress—otherwise we risk doing duplicate or conflicting work. But if another instance did move forward, what’s the right course of action?
We also thought about doing the re-subscription from the ResilientGrpcSubscription, but since that’s part of the network layer, calling factus.subscribe from there feels like mixing concerns.

2. Once the instance releases the token, it’s game over.
There’s no mechanism to re-acquire it. The subscription is effectively dead, and the projection stays unhealthy with no recovery path—unless the container is rotated and a fresh instance starts from scratch.

We should discuss if there’s a better solution we haven’t thought of yet. Otherwise, it feels like the effort needed here might far outweigh the actual value we’d get from solving it.

uweschaefer:
After discussing in several 1:1 meetings, there does not seem to be an easy solution.
We'd need to go back to the whiteboard to plan a (probably major) structural change.
Until then, it'd be best to postpopne the ticket.

PR #3401, opened 2025-01-14
#2841: Invalidate writer token on close

mamup11:
closes: #2841


uweschaefer:
RLock object behaves according to the Java Lock specification. It means only lock owner thread can unlock it otherwise IllegalMonitorStateException would be thrown. Otherwise consider to use [RSemaphore](https://redisson.org/docs/data-and-services/locks-and-synchronizers/#semaphore) object.

otbe (approved):

otbe:
For me it looks fine but I would still be interested in what @uweschaefer thinks about it  :)

uweschaefer (approved):
thx @mamup11

PR #3403, opened 2025-01-14
#2531 cleanup notification handling

### LINKED ISSUE #2531: In readonly mode it should be possible to work with a hot standby DB
With the recently introduction of the [readOnlyMode](https://github.com/factcast/factcast/pull/2334) it is possible to run a factcast server which is able to answer queries but does not write to database. This includes obvious things like fact publication but also scheduled tasks like tail index management and schema registry updates.

It would be nice to be able to run this factcast server against a hot standby (like an RDS read replica) in order to offload work from the primary. However this is not possible right now. Internally we use [LISTEN](https://www.postgresql.org/docs/9.1/sql-listen.html) for certain things like being notified about a new fact published for a certain active subscription or getting notified about a new blacklisted fact. `LISTEN` is [not possible](https://www.postgresql.org/docs/current/hot-standby.html) with a hot standby.

We should think about alternatives or at least document this limitation of the readOnlyMode.

uweschaefer:
prerequisite for #2531


factcast-store/src/main/java/org/factcast/store/internal/filter/blacklist/PgBlacklistDataProvider.java
```diff
@@ -46,16 +44,17 @@ public PgBlacklistDataProvider(
   @Override
   public void afterSingletonsInstantiated() {
     bus.register(this);
-    updateBlacklist();
+    updateBlacklist(); // just to be safe
```
otbe, line 47: Actually I think its necessary otherwise the readonly instance would operate on an empty blacklist until the first change notification comes in, isnt it?

factcast-store/src/main/java/org/factcast/store/internal/listen/PgListener.java
```diff
@@ -115,138 +113,92 @@ private void connectionSetup(PgConnection pc) throws SQLException {
...
-            n.getName());
-        postFactInsertionSignal(PgConstants.CHANNEL_FACT_INSERT);
-      }
-    }
-  }
-
-  private void postBlacklistChangeSignal() {
-    log.trace("Potential blacklist change detected");
-    eventBus.post(new BlacklistChangeSignal());
+    List<PGNotification> list = List.of(notifications);
```
mamup11, line 142: Just commenting that this is a great simplification of the previous code, great to see!

PR #3455, opened 2025-02-12
OpenRewrite code cleanups

uweschaefer:
very minor ones from org.openrewrite.staticanalysis.CodeCleanup

It already broke a few things:
* Hide constructor for Main classes
* lots of things in kotlin code (reverted all of those changes)

rest looks reasonable 
* mandatory usage of {} with if/then/else
* import cleanups / FQCN references
* yoda coding on equals
* field initialization with default values


uweschaefer:
That's a subtle one :)
image

uweschaefer:
Don't even look at codecov.

factcast-core/src/main/java/org/factcast/core/FactHeader.java
```diff
@@ -27,9 +27,9 @@
 @EqualsAndHashCode(of = {"id"})
 public class FactHeader {
 
-  @JsonProperty @NonNull UUID id;
+  @JsonProperty UUID id;
 
-  @JsonProperty @NonNull String ns;
```
mamup11, line 32: I wonder, why does it remove the @NonNull here?

factcast-factus/src/main/java/org/factcast/factus/lock/InLockedOperation.java
```diff
@@ -42,4 +42,6 @@ public static void assertNotInLockedOperation() {
               + " allowed");
     }
   }
+
+  private InLockedOperation() {}
```
mamup11, line 46: In some other class, it added a @UtilityClass? in this one it added a private constructor? and I think in the main class it added a NoArgsConstructor, not sure if those were made by the script or your changes directly, but just wondering about it

factcast-server-ui/package.json
```diff
@@ -3,83 +3,33 @@
   "dependencies": {
     "@polymer/polymer": "3.5.2",
     "@vaadin-component-factory/vcf-popup": "24.0.3",
-    "@vaadin/a11y-base": "24.5.1",
```
mamup11, line 6: This is a huge change in the dependencies, was this expected?

factcast-factus/src/main/java/org/factcast/factus/lock/InLockedOperation.java
uweschaefer: good catch. mosty, where the script added private constr. i either changed it to public, or made it a util class. will change that

factcast-server-ui/package.json
uweschaefer: not at all. that is unrelated from running ui server

factcast-core/src/main/java/org/factcast/core/FactHeader.java
uweschaefer: because the field is initialized with null.

factcast-store/src/main/java/org/factcast/store/internal/CondensedQueryExecutor.java
```diff
@@ -122,14 +122,14 @@ public void onEvent(FactInsertionNotification ev) {
   @VisibleForTesting
   boolean mightMatch(String ns, String type) {
     return ns == null
-        ||
         // listens to this exact type
```
leflamm, line 125: not really important but it kinda duplicated the comments... 

image


```suggestion
```

factcast-store/src/main/java/org/factcast/store/internal/CondensedQueryExecutor.java
```diff
@@ -122,14 +122,14 @@ public void onEvent(FactInsertionNotification ev) {
...
+        || // listens to this exact type
         interests.contains(ns + ":" + type)
-        ||
         // listens to the whole namespace
```
leflamm, line 128:
```suggestion
```

factcast-store/src/main/java/org/factcast/store/internal/CondensedQueryExecutor.java
```diff
@@ -122,14 +122,14 @@ public void onEvent(FactInsertionNotification ev) {
...
+        || // listens to the whole namespace
         interests.contains(ns)
-        ||
         // is a catchall
```
leflamm, line 131:
```suggestion
```

leflamm (commented):
a minor comment added (in 3 parts :slightly_smiling_face: )

leflamm (approved):

PR #3476, opened 2025-02-25
#2841: Retry get token

### LINKED ISSUE #2841: [Factus] A WriterToken that gets invalid during normal operations leads to a dead subscription

mamup11:
closes #2841


factcast-factus/src/main/java/org/factcast/factus/FactusImpl.java
```diff
@@ -176,17 +192,28 @@ private <P extends SubscribedProjection> Subscription doSubscribe(
         new AbstractFactObserver(subscribedProjection, PROGRESS_INTERVAL, factusMetrics) {
 
           FactStreamPosition lastPositionApplied = null;
+          WriterToken subscriptionToken = token;
 
           @Override
           public void onNextFacts(@NonNull List<Fact> elements) {
-            assertTokenIsValid();
+            reAcquireTokenIfNeeded();
```
uweschaefer, line 199: i may be wrong, but simply retrying here might be a catastrophe waiting to happen:

Scenario: 
* Instance A has the token, continously processes messages
* Token expires. On the next message this is called, trying to reacquire
* In the mean time instance B has taken over and continues applying, even far beyond what 'elements` contain.
* At some point (maybe days later) B is killed.
* Now A takes over still waiting to apply 'elements' which is 
 a) already applied
 b) screwing the order of event applications
 c) resetting the position to "few days ago", effectively re-applying everything after 'elements'.

please tell me that i am wrong.

mamup11: That’s sadly true, that scenario was not covered in the int test but it’s true that it would behave like that. Then the refresh of the token needs to happen some other way, I will think about iy

PR #3503, opened 2025-03-05
3502: Fixing updating date pickers from pre filled url

mamup11:


factcast-server-ui/src/main/java/org/factcast/server/ui/full/FullQueryBean.java
```diff
@@ -30,7 +30,6 @@
 @Data
 @SuppressWarnings("java:S1948")
 @Accessors(fluent = false, chain = false)
-@JsonIgnoreProperties("since")
```
uweschaefer, line 33: since & until should be excluded from the url to avoid confusion

uweschaefer:
thx @mamup11

PR #3514, opened 2025-03-10
#3513: Fix equals and hashCode for FactSpecs

### LINKED ISSUE #3513: Equals generated by Lombok fails for FactSpecs from old and new clients
https://github.com/factcast/factcast/blob/65ec70cba5c6cdc61111bf1a07e0b3b3c85d6832/factcast-core/src/main/java/org/factcast/core/spec/FactSpec.java#L33

Lombok does not use the mergedAggIds getter as of{“““} expects fieldnames rather than propertynames.

## Status Quo:

either a factSpec comes from an older client and has onlky an “aggid” set, or

it comes from a recent client that has only aggIds set, no matter if it is none/one/many

## Goal: 

we want to be equal (incl. hashcode) for the two factSpecs(aggid=7) and factSpec(aggIds=[7]).

But not equal for factSpecs(aggid=7) and factSpec(aggIds=[8]).

bedaka:
closes #3513


uweschaefer (approved):

uweschaefer:
thx @bedaka

PR #3518, opened 2025-03-11
#issue3131: warn in case of misconfiguration

bedaka:


factcast-server-ui-s3-reportstore/src/main/java/org/factcast/server/ui/s3reportstore/S3ReportStoreConfiguration.java
```diff
@@ -43,10 +46,17 @@ public S3TransferManager s3TransferManager(S3AsyncClient client) {
...
   }
 
+  @Value("${factcast.ui.report.store.path:}")
+  private String filesystemReportPath;
+
   @Bean
   @ConditionalOnMissingBean
   ReportStore s3ReportStore(
       S3AsyncClient s3Client, S3TransferManager s3TransferManager, S3Presigner s3Presigner) {
+    if (!filesystemReportPath.isEmpty()) {
```
uweschaefer, line 56: Just to make sure: isn't that NPE prone, even though ":" was used at the end of the expression?

factcast-server-ui-s3-reportstore/src/test/java/org/factcast/server/ui/s3reportstore/config/AwsS3TestConfig.java
```diff
@@ -33,16 +33,18 @@
 @Configuration
 public class AwsS3TestConfig {
 
+  @Value("${factcast.ui.report.store.s3}")
```
uweschaefer, line 36: that should be replaced by using the Properties instance, right?

factcast-server-ui-s3-reportstore/src/main/java/org/factcast/server/ui/s3reportstore/S3ReportStoreConfiguration.java
bedaka: removed the unnecessary config as discussed

factcast-server-ui-s3-reportstore/src/test/java/org/factcast/server/ui/s3reportstore/config/AwsS3TestConfig.java
bedaka: removed the unnecessary config as discussed

uweschaefer:
thx @bedaka

PR #3532, opened 2025-03-12
Documentation aggids

uweschaefer:


factcast-site/documentation-docsy/content/en/Concept/factstreams.md
```diff
@@ -44,23 +52,97 @@ Here the second Fact obviously has impact on both the following, as well as the
...
+Consider an Aggregate User (pseudocode):
+
+```
+User {
+ UUID id
+ String name
+}
+```
+
+and Facts like `UserCreated` (skipping here), and `UserDeleted` containing just the id of the user to delete.
```
bedaka, line 82: seems like there is something missing in `(skipping here)`

factcast-site/documentation-docsy/content/en/Concept/factstreams.md
```diff
@@ -44,23 +52,97 @@ Here the second Fact obviously has impact on both the following, as well as the
...
+
+This will probably lead to an Aggregate class User with handler methods similar to: (using Factus Syntax here for
```
bedaka, line 84:
```suggestion
This will lead to an Aggregate class User with handler methods similar to: (using Factus Syntax here for
```

factcast-site/documentation-docsy/content/en/Concept/factstreams.md
```diff
@@ -44,23 +52,97 @@ Here the second Fact obviously has impact on both the following, as well as the
...
+But what happens, if you add another user reference to the UserDeleted Event?
+
+```
+UserDeleted {
+ UUID id // the one to actually delete
+ UUID deletedByUserId // the user that triggered the deletion
+}
+```
+
+If the handlers stay unchanged, then (as the `aggIds` should contain all the aggregates involved)
```
bedaka, line 116: Feel like this should read:
```suggestion
If the handlers stay unchanged, then (as the `aggIds` will contain all the aggregates involved)
```

factcast-site/documentation-docsy/content/en/Concept/factstreams.md
```diff
@@ -44,23 +52,97 @@ Here the second Fact obviously has impact on both the following, as well as the
...
+one Fact would trigger not only the intended deletion, but also the deletion of the user triggering.
```
bedaka, line 117: Just a suggestion to maybe phrase "trigger the deletion" more explicit:

[...] one UserDeleted Fact would not only be consumed in the Aggregate of the deleted user but also in the Aggregate of the user that triggered the deletion, resulting in both being marked as deleted.

factcast-site/documentation-docsy/content/en/Concept/factstreams.md
uweschaefer: so i'd stick to should here

bedaka: Makes sense 👍

bedaka (approved):

PR #3589, opened 2025-04-05
Revert "fix(deps): update dependency org.apache.fury:fury-core to v0.10.0"

uweschaefer:
Reverts factcast/factcast#3461

as the test shows (and is supposed to show), 0.10 is not compatible with versions before. we can only do this in a major release.


bresciamattia (approved):
thanks for noticing it :)

Issue #3449, opened 2025-02-11
Migrate schemareg-cli to spring


uweschaefer:
We used micronaut for the CLI with the hope to be able to compile down to a native binary. This was not entirely possible at the time.
In an attempt to unify the use of frameworks it'd be nice if we'd migrate it to spring.

As frameworks have improved their way of creating native executables quite a lot, we might retry going native from there.

PR #3614, opened 2025-04-16
#3449: schema-reg-cli migration to spring

uweschaefer:


leflamm (approved):
Nothing to say here :-)

Issue #3612, opened 2025-04-16
Feature: Evict respective transformed versions for updated facts


uweschaefer:
Even though strongly discouraged, when people jump in and change the past (by updating events), we need to make sure, that no transformed (cached) versions survive, that were based on the fact before update.
A simplistic approach would be to evict **all** cached versions of an updated fact using a trigger.

Edge use-case, but might help.

PR #3620, opened 2025-04-22
#3612: Evict transformation cache when fact is updated

bedaka:
Updating facts in general is not recommended as it breaks with the immutability of the fact stream. In cases where it is required we want to make sure that we're not dealing stale data after the update by providing cached transformations of an updated fact to the client. 

Therefore a trigger is added on the fact table for all UPDATE operations, which causes a removal of transformations for that factId from the cache.

### DIFF
--- a/factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
+++ b/factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
@@ -107,28 +107,32 @@ public class InMemTransformationCache implements TransformationCache {
   }
 
   @Override
-  public synchronized void invalidateTransformationFor(String ns, String type) {
-    Set<Key> toBeInvalidated =
-        cache.entrySet().stream()
-            .filter(
-                e ->
-                    e.getValue().fact().ns().equals(ns)
-                        && Objects.equals(e.getValue().fact().type(), type))
-            .map(Entry::getKey)
-            .collect(Collectors.toSet());
-    if (!toBeInvalidated.isEmpty()) {
-      toBeInvalidated.forEach(cache::remove);
+  public void invalidateTransformationFor(String ns, String type) {
+    synchronized (cache) {
+      Set<Key> toBeInvalidated =
+          cache.entrySet().stream()
+              .filter(
+                  e ->
+                      e.getValue().fact().ns().equals(ns)
+                          && Objects.equals(e.getValue().fact().type(), type))
+              .map(Entry::getKey)
+              .collect(Collectors.toSet());
+      if (!toBeInvalidated.isEmpty()) {
+        toBeInvalidated.forEach(cache::remove);
+      }
     }
   }
 
   @Override
-  public synchronized void invalidateTransformationFor(UUID factId) {
-    Set<Key> toBeInvalidated =
-        cache.keySet().stream()
-            .filter(e -> e.id().contains(factId.toString()))
-            .collect(Collectors.toSet());
-    if (!toBeInvalidated.isEmpty()) {
-      toBeInvalidated.forEach(cache::remove);
+  public void invalidateTransformationFor(UUID factId) {
+    synchronized (cache) {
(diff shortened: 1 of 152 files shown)


mamup11 (commented):
Maybe there should be also an int test, I saw PgListenerIntegrationTest, perhaps something similar would be useful?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgFactUpdateListener.java
```diff
@@ -0,0 +1,85 @@
...
+import org.factcast.store.registry.transformation.chains.TransformationChains;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+
+@RequiredArgsConstructor
+@Slf4j
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class PgFactUpdateListener implements SmartInitializingSingleton, DisposableBean {
```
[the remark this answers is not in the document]
bedaka: right, good catch totally forgot about that.

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
```diff
@@ -121,6 +121,17 @@ public synchronized void invalidateTransformationFor(String ns, String type) {
     }
   }
 
+  @Override
+  public synchronized void invalidateTransformationFor(UUID factId) {
```
uweschaefer, line 125: sync seems wrong here, in order to avoid CMEs, we'need to sync on the cache itself (being a sync-collection).
This bug already existed in the other invalidate method.

fixed it.

factcast-store/src/test/java/org/factcast/store/internal/listen/PgListenerIntegrationTest.java
```diff
@@ -167,4 +169,108 @@ public void reset() {
...
+      // RUN
+      final var statement =
+          pc.prepareStatement(
+              String.format(
+                  "UPDATE fact SET payload = '{\"value\":\"2\"}' WHERE header @> '{\"type\": \"%s\"}'",
+                  "listenerTest1"));
+      statement.executeUpdate();
+
+      // ASSERT
+      // first, check trigger
```
uweschaefer, line 234: we're testing something different then what the code does, here. The listener assumes there was an "id" field in your notification, while here we're testing "header.id" - will change the test and fix the resulting bug .

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgFactUpdateListener.java
```diff
@@ -0,0 +1,85 @@
...
+  private final TransformationCache cache;
+
+  private final TransformationChains chains;
+
+  private ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+
+  static final long INFLIGHT_TRANSFORMATIONS_DELAY_SECONDS = 10L;
+
+  @VisibleForTesting
+  protected PgFactUpdateListener(
```
leflamm, line 49: used anywhere?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgFactUpdateListener.java
```diff
@@ -0,0 +1,85 @@
...
+@RequiredArgsConstructor
+@Slf4j
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class PgFactUpdateListener implements SmartInitializingSingleton, DisposableBean {
+
+  private final EventBus bus;
+
+  private final TransformationCache cache;
+
+  private final TransformationChains chains;
```
leflamm, line 42: used for anything?

leflamm (commented):
@bedaka just a minor question about seemingly unused code

PR #3696, opened 2025-05-08
3603 Unify DynamoManagedUserNames

mamup11:


factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/DynamoITest.java
```diff
@@ -300,7 +300,7 @@ public WriterToken acquireWriteToken(@NonNull Duration maxWait) {
     }
 
     @Override
-    protected void apply(UserCreated created) {
+    public void apply(UserCreated created) {
```
uweschaefer: handler methods should not be public, we even print a warning there.
Why is it necessary?

uweschaefer: changed it

uweschaefer:
lgtm, thx @mamup11

factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/DynamoITest.java
uweschaefer: AH! because you pulled them in the interface, thus making them implicitely public.
Sorry, did not get that.

PR #3718, opened 2025-05-20
Issue2645: Add enumerate versions and enhance version selection in UI

bedaka:


bedaka:
TODO: adding / adapting the index for the version queries.

factcast-store/src/main/resources/db/changelog/factcast/issue2645/create_published_versions_for_existing_events.sql
```diff
@@ -0,0 +1,12 @@
+-- add an entry with the first and last ser for every day
+-- this should be run manually if you have a big fact table already,
+-- as it can take some time depending on your DB performance
+
+INSERT INTO publishedschemaversions
```
uweschaefer, line 5: select/group seems to be more efficient

```
postgres=> insert into schema_mig_perf_test  ( select header ->> 'ns' as ns, header ->>'type' as type, ((header ->>'version'))::int as version from fact group by ns,type,version)
;
INSERT 0 791
Time: 7894.467 ms (00:07.894)
postgres=> truncate schema_mig_perf_test                                                                                                                                          
;
TRUNCATE TABLE
Time: 5.474 ms
postgres=> INSERT INTO schema_mig_perf_test
    (SELECT header ->> 'ns'   AS ns,
            header ->> 'type' AS type,
            cast((header ->> 'version') AS INT)
     FROM fact)
ON CONFLICT DO NOTHING;
INSERT 0 791
Time: 9054.902 ms (00:09.055)
```

The diff is not huge, but might be worth it on huge fact tables

factcast-store/src/main/resources/db/changelog/factcast/issue2645/create_published_versions_trigger.sql
```diff
@@ -0,0 +1,34 @@
+-- on fact insert we're saving the type and version of the fact to
+-- prevent extensive queries when enumerating them later.
+
+CREATE TABLE IF NOT EXISTS publishedschemaversions
```
uweschaefer, line 4:
```suggestion
CREATE TABLE IF NOT EXISTS published_schema_versions
```
for readability & consistency

uweschaefer: would need to be adapted of course

factcast-store/src/main/resources/db/changelog/factcast/issue2645/create_published_versions_for_existing_events.sql
uweschaefer: ON CONFLICT DO NOTHING; would still have to be added, just in case

factcast-store/src/main/resources/db/changelog/factcast/issue2645/create_published_versions_trigger.sql
```diff
@@ -0,0 +1,34 @@
...
+(
+    ns      varchar(255),
+    type    varchar(255),
+    version int,
+    UNIQUE (ns, type, version)
```
leflamm, line 9: do you think at some point we'd benefit from an index fitting the main access pattern `SELECT_DISTINCT_VERSIONS_FOR_NS_AND_TYPE` ?

bedaka: we agreed to leave this out for now

Issue #3829, opened 2025-07-02
Scheduling future events (low priority)


BernhardBln:
**Is your feature request related to a problem? Please describe.**

Sometimes an aggregate’s lifecycle includes events that must appear in the event stream at a precise time to trigger a state transition exactly then.

This helps:

✅ Reject conflicting events issued after the scheduled one (using optimistic locking)
✅ Ensure business rules are respected

⚠️ There are already good enough patterns for this, and I don’t expect to need this anytime soon. However, these patterns have some drawbacks, and I found it frustrating that we couldn’t come up with a cleaner solution. See the alternatives section below.

An example is a simple auction lifecycle: it starts with StartAuction, followed by multiple BiddingAccepted events, then AuctionFinished to complete the process. Additionally, AuctionCancelled can be used to cancel the auction before it finishes, and AuctionInvalidated to invalidate it after it has already finished.

Usually, this is handled by a job in your service that publishes AuctionFinished exactly when the auction ends. The drawback: under heavy load or in case of other issues, the event might be published too late, violating business rules (and looking weird, as the page still displays the auction as open).
Alternatively, clients who know the auction end time compare it to their local clocks—which may differ—causing situations where one bidder’s offer, though placed before another bidder’s, is rejected, while the other offer is still accepted.

**Describe the solution you'd like**

**Idea:**

One idea would be to use postgres as the time authority, as we usually have just one instance here. Then we could solve the problem by allowing to schedule events (and, if necessary, to unschedule/cancel them again). Scheduling and unscheduling should also be possible inside of transactions.

Then you could do something like this inside a transaction (and outside as well):

```java
factus.withLockOn(AuctionAggregate.class, auctionId)
  .attempt((auction, tx) -> {

    if (auction.isNotCancelledOrFinished()){
      tx.publish(auctionStarted);
      tx.schedule(AuctionFinished, startAuction.getEndTime(), auction.getId());
    }

  });
```

The method `auction.isNotCancelledOrFinished()` would **only** rely on the presence of events such as AuctionFinished, and **not** use the time of the auction end and compare it with the local time.

Then we either have startAuction published, AND scheduled that AuctionFinished is published at the end time of the auction, or both did not happen.

Then AuctionFinished will be automatically inserted into the event stream at the given time (with the postgres as the time authority), hence also appearing in subscribed projections, unless it is unscheduled before that time.

We have two requirements here:

[...]

PR #3843, opened 2025-07-11
#3829 added clock to factus

BernhardBln:
Only disadvantage - this relies on a method in the factus interface that is annotated as "subject to change, use at your own risk"...

### DIFF
--- a/factcast-factus/src/main/java/org/factcast/factus/Factus.java
+++ b/factcast-factus/src/main/java/org/factcast/factus/Factus.java
@@ -20,6 +20,7 @@ import com.google.common.cache.*;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import java.io.Closeable;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.function.*;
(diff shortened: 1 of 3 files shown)


factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
```diff
@@ -886,6 +887,28 @@ private static void sleep(long ms) {
...
+      long beforeMillis = System.currentTimeMillis();
+      long now = factus.currentTimeInMillis();
+      long afterMillis = System.currentTimeMillis();
+
+      assertThat(now).isGreaterThanOrEqualTo(beforeMillis).isLessThanOrEqualTo(afterMillis);
+    }
+
+    @Test
+    void currentTimeAsInstant() {
+      Instant before = Instant.now();
```
uweschaefer, line 904: bit flaky, isn't it? This relies on local clock and PG clock being reasonably in sync.

factcast-factus/src/main/java/org/factcast/factus/Factus.java
```diff
@@ -220,4 +221,24 @@ default <P extends SubscribedProjection> void waitFor(
...
   @NonNull
   FactStore store();
+
+  /**
+   * @return Current time in seconds since 1970-01-01 00:00:00-00 (can be negative) from the
+   *     factstore.
+   *     <p>Can be used to synchronize across clients (e.g. to decide if it is too late to publish a
+   *     certain event)
+   */
+  default long currentTimeInMillis() {
```
uweschaefer, line 231: Think we can drop this one. the extra effort of wrapping into Instant is insignificant.

uweschaefer (changes requested):

factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
BernhardBln: Sorry for the late response, I think I need to turn on github notifications again :D 

I don't think so, in this test we use the pg from a docker container, and AFAIK their time would be the same as the host that is running the tests.

PR #3886, opened 2025-07-30
lock namespaces

### LINKED ISSUE #3877: Check if throughput can be increased if we only lock namespaces on inserts
Just wondering, if we wouldn't serialise inserts but only lock namspaces that are affected by our operation (simple inserts: the namespace of the event(s) that get inserted, locked inserts: the namespaces of all factspecs of the locked projection + the namespace of the event to insert), if this could significantly increase throughput for parallel writes in distinct namespaces. 

Create a table `namespaces(namespace varchar primary key, code smallserial)`, on startup or when syncing schema registry, insert all namespaces that are still missing, but do not touch the existing entries ever, and have a loading cache that gets the codes from there.

when locking, don't just get one advisory lock, but do something like

```java
// for simple inserts
copiedListOfFacts.stream()
    .map(Fact::ns)
    .distinct()
    .map(codeCache::get)
    .sorted()
    .forEachOrdered(advisoryLock::aquire);

// for locked 
Stream.concat(
    factsToPublish.stream().map(Fact::ns),
    state.get().specs().stream().map(FactSpec::ns)
). distinct()
    .map(codeCache::get)
    .sorted()
    .forEachOrdered(advisoryLock::aquire);
```

**As a first test, we can also just disable getting the advisory lock and see if that makes a difference already.**

A different data store 😆

BernhardBln:
closes #3877

### DIFF
--- a/factcast-store/src/main/java/org/factcast/store/internal/PgFactStore.java
+++ b/factcast-store/src/main/java/org/factcast/store/internal/PgFactStore.java
@@ -15,29 +15,52 @@
  */
 package org.factcast.store.internal;
 
+import static org.factcast.store.internal.lock.FactTableWriteLock.STAR_NAMESPACE_CODE;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Lists;
-import java.sql.*;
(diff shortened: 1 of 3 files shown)


uweschaefer:
turns out that other factors are limiting way before the lock scope could be. (NOTIFY/date2serial,...) closing for the foreseeable future.

Issue #3904, opened 2025-08-06
Performance bug: create state token


uweschaefer:
**Describe the bug**
In order to retrieve the latest statetoken for a list of specs, or for an aggregate, we currently use:

```sql
select ser from fact where ... ORDER BY ser DESC limit 1;
```

This is basically a backward scan over pkey

```
postgres=> explain select ser from fact where header@>'{"ns":"X"}' ORDER BY ser DESC limit 1;
                                           QUERY PLAN                                           
------------------------------------------------------------------------------------------------
 Limit  (cost=0.57..1683.78 rows=1 width=8)
   ->  Index Scan Backward using fact_pkey on fact  (cost=0.57..20250747.88 rows=12031 width=8)
         Filter: (header @> '{"ns": "X"}'::jsonb)
(3 rows)
```

There are different things to ponder:
* While this might seem clever, because it just scans from the tail of the stream and stops as soon as it found something, if we have a reasonable amount of facts not matching, is pretty wasteful.
* if the client already knows some factstreamposition matching (by looking at a projection or aggregate) this could help adding a ser>X criterion and pg might switch to a plan that involves the proper index for filtering (maybe even a tail)
* we could actually FORCE using a plan that involves the correct index by using WITH (subqueries wont work as they are still optimized together)

like 
```sql
postgres=> explain analyze WITH subq as (select ser from fact where header@>'{"ns":"X"}' and ser>115000000 ) select max(ser) from subq;                                                                                                
                                                               QUERY PLAN                                                     
           
------------------------------------------------------------------------------------------------------------------------------
-----------
 Aggregate  (cost=12285.30..12285.31 rows=1 width=8) (actual time=93.437..93.439 rows=1 loops=1)
   ->  Bitmap Heap Scan on fact  (cost=77.44..12283.91 rows=557 width=8) (actual time=27.304..92.903 rows=10357 loops=1)
         Recheck Cond: (header @> '{"ns": "X"}'::jsonb)
         Filter: (ser > 115000000)
         Rows Removed by Filter: 53093
         Heap Blocks: exact=53489
         ->  Bitmap Index Scan on idx_fact_header  (cost=0.00..77.30 rows=12031 width=0) (actual time=15.848..15.849 rows=6345
0 loops=1)
               Index Cond: (header @> '{"ns": "X"}'::jsonb)
 Planning Time: 0.179 ms
 Execution Time: 94.167 ms
(10 rows)

```

which gets even better on the tail indexes.

[...]

Issue #3900, opened 2025-08-04
Prepare for pluggable locking strategies


uweschaefer:
**Is your feature request related to a problem? Please describe.**
There are a few ideas of how to improve the write throughput of the factstore.
Here we extract the relevant code into a strategy and provide a benchmark.

PR #3907, opened 2025-08-07
#3900: draft insert concurrency strategies

uweschaefer:
refactored a bit, and added two concurrency strategies next to the current (legacy).
You can toggle and test by switching 'factcast.store.concurrencyStrategy'.

As it turns out, locking is not an issue in an realistic benchmark (criticism welcome).

I tried a scenario with 64 threads inserting 10k facts with 20% of them added conditionally (with 20ms extra latency between getting and using a token).

When executing against a standalone postgres, we found two much more limiting factors to write throughput than the locking itself:

#3911 #3910 

But there is an interesting resilience aspect in here: when using the new strategies, a test for conflicts is no longer part of the lock scope (either not blocking unconditional inserts, or all, depending on the strategy). In cases where those tests take up much more time than expected, the server is not blocked any more.

Also there were some findings about `getStateFor`that lead to #3904.

While this might be helpful, the main boost will not be visible before we tackled #3911 and #3910.

I'd suggest to test if the UNLOCKED_CHECK Strategy is safe&ready to become the new default at some point.


uweschaefer:
caused by #3886

uweschaefer:
The UNLOCKED_CHECK is currently **NOT** safe to use.

It breaks a fundamental assumption that is currently used everywhere, in follow mode, catchup etc.
This assumption is: as soon, as i see serial X, X-n everything before does not change.
For instance, when doing a catchup we query the fact table until its end, and in the follow mode, we query< from there.
While this works often times, there is a serious bug introduced by any relaxation of the strict serialiazation of inserts: older (ser<the current one) might become visible afterwards.

Consider the stream having 4 concurrent inserts 1,2,3,4. As there is no defined order any longer, the might arrive in the order 1,2,4,3. Considering all of them being relevant to a subscription, it might happen, that at the time of query it sees 1,2,4. This will lead to the assumption that following, we can query WHERE ser>4. 
Here the subscription never sees fact 3.

uweschaefer:
obsolete with #4745

Issue #3917, opened 2025-08-12
Reduce duplicate log lines


bedaka:
At the moment our factstore frequently hits the rate limit of the logging infrastructure, resulting in delayed logs and false positive "no logs ingested" alerts.

In combination with other measures like adjusting the batch size of the otel collector, it might make sense check for duplicate log lines and unify them.

PR #3918, opened 2025-08-12
#3917-duplicate-log-lines: issue3917: remove duplicate log lines

bedaka:


factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -74,7 +74,7 @@ public class PgFactStream {
 
   void connect(@NonNull SubscriptionRequestTO request) {
     this.request = request;
-    log.debug("{} connect subscription {}", request, request.dump());
```
uweschaefer, line 77: is that really intended? 
req.dump() was supposed to be useful for debug purposes, where req.toString() is only the name.

PR #3936, opened 2025-08-20
#3935: fix NPE

### LINKED ISSUE #3935: NPE can be triggered when factStream is closed before connecting
**Describe the bug**
A NPE can happen if for whatever reason the PgFactStream gets closed before it managed to connect and therefore create a request.

bresciamattia:
closes #3935


uweschaefer:
ready for review or still draf @bresciamattia ?

PR #4070, opened 2025-10-19
InLockedOperation for virtual threads - fixes #4064

### LINKED ISSUE #4064: InLockedOperation uses ThreadLocal
Factus uses a threadlocal in InLockedOperation to prevent accidental use of factus (instead of tx) in a locked operation. That will not work any more when a client has virtual threads enabled.

(Not sure it's the only problem)

Can probably be fixed using ScopedValue. I can prepare an MR if desired.

BernhardBln:
Ok done here and seems to be working, but a bit annoying at the moment.

The sub module needs to be build with jdk 25. But jdk 25 has issues with lombok at the moment, it seems (although I thought it was fixed already).

So currently you need to do the following to build the project:

- run `mvn clean install` in `factcast-factus-jdk25`, with JAVA_HOME pointing at a JDK 25
- then in the project root, run with JAVA_HOME pointint at a JDK 17: `mvn install -pl !factcast-factus-jdk25`

Shall we create a separate job in the build pipeline? No clue how complicated that is.

The same problem we would have if we added an integration test for that. 

I tested this manually for now and seems to work:

image

Should I also add an info to the root Readme?

### DIFF
--- a/factcast-factus/src/test/java/org/factcast/factus/lock/LockedUtilTest.java
+++ b/factcast-factus/src/test/java/org/factcast/factus/lock/LockedUtilTest.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2017-2025 factcast.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.factcast.factus.lock;
+
+import static org.assertj.core.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class LockedUtilTest {
+
+  @Test
+  void isVirtualThreadSupported() {
+    assertThat(LockedUtil.isVirtualThreadSupported()).isTrue();
+  }
+
+  @Test
+  void isScopedValueAvailable() {
+    assertThat(LockedUtil.isScopedValueAvailable()).isTrue();
+  }
+}


factcast-factus-jdk25/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>${maven.compiler.version}</version>
+        <configuration>
+          <source>${maven.compiler.source}</source>
+          <target>${maven.compiler.target}</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
```
BernhardBln: probably not needed any more.

factcast-factus/src/test/java/org/factcast/factus/FactusImplTest.java
```diff
@@ -471,8 +486,10 @@ void withLockOnAggregateClass_aggregateExists() {
       when(fc.subscribe(any(), any()))
           .thenAnswer(
               inv -> {
-                // There must be one fact referencing this aggregate to make sure the
-                // fact position is set, so that we do not get an empty optional when doing find.
+                // There must be one fact referencing this aggregate to make sure
```
uweschaefer: why?

factcast-factus-jdk25/src/main/java/org/factcast/spring/boot/autoconfigure/factus/FactusJdk25AutoConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+import org.factcast.factus.lock.InLockedOperation;
+import org.factcast.factus.lock.InLockedOperationForVirtualThreads;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureOrder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.Ordered;
+
+@AutoConfiguration
+// needs to come before FactusAutoConfiguration
+@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE - 1)
```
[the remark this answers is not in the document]
uweschaefer: we all do :)

PR #4075, opened 2025-10-22
Issue3903 Solve EventExplorer going OOM

### LINKED ISSUE #3903: Server UI Report generation can run into OOM
**Describe the bug**
This needs some investigation, as the reporting feature is meant for – well reporting –  it should be able to handle big amounts of data by streaming it.

If it already does, then we should probably add some advice to the docs on how to configure it (grpc buffer size, instance mem min, etc)

bedaka:
Closes #3903

### DIFF
--- a/factcast-server-ui/src/main/java/org/factcast/server/ui/port/ReportUploadStream.java
+++ b/factcast-server-ui/src/main/java/org/factcast/server/ui/port/ReportUploadStream.java
@@ -28,7 +28,6 @@ import org.factcast.server.ui.report.ReportFilterBean;
 @Slf4j
 public abstract class ReportUploadStream {
   @NonNull private final JsonGenerator jsonGenerator;
-  private final String reportName;
 
   @SneakyThrows
   protected ReportUploadStream(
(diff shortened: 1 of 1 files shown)


factcast-server-ui-s3-reportstore/src/test/java/org/factcast/server/ui/s3reportstore/S3ReportStoreTest.java
```diff
@@ -72,40 +69,36 @@ class WhenSavingReports {
...
-      final var fileUpload = mock(Upload.class);
-      final var completedUpload = mock(CompletedUpload.class);
-      when(s3TransferManager.upload(any(UploadRequest.class))).thenReturn(fileUpload);
-      when(fileUpload.completionFuture())
-          .thenReturn(CompletableFuture.completedFuture(completedUpload));
       objectDoesNotExist();
+      final var completedUpload = mock(CreateMultipartUploadResponse.class);
+      when(s3Client.createMultipartUpload(any(CreateMultipartUploadRequest.class)))
+          .thenReturn(CompletableFuture.completedFuture(completedUpload));
+      // TODO: why no check?
```
leflamm: can go?

PR #4185, opened 2025-12-01
#4127: catch anything and end the subscription

### LINKED ISSUE #4127: Missing exception handling causes skipped events for subscribed projections in follow phase
Let's say we have a subscribed projection that's in the follow phase. 
At some point the `CondensedQueryExecutor` will execute the [run](https://github.com/factcast/factcast/blob/e23fd28e678693c70432be5eaf14d900dd1e0959/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java#L95) method of the `PgSynchronizedQuery`. 

Let's also assume we only process one fact that needs transformation during this execution. The [row callback handler](https://github.com/factcast/factcast/blob/e23fd28e678693c70432be5eaf14d900dd1e0959/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java#L139) is going to pass the one fact into the pipeline where it is buffered due to the need of transformation. However the serial is [already moved](https://github.com/factcast/factcast/blob/e23fd28e678693c70432be5eaf14d900dd1e0959/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java#L153). Because we only consume one event we move on to the finally block of the `run` method and trigger a [flush](https://github.com/factcast/factcast/blob/e23fd28e678693c70432be5eaf14d900dd1e0959/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java#L123) of the pipeline. 
This triggers at the end also the transformation.

If an exception happens that is not handled in between this will cause a skipping of events because no facts are sent to the client but also no error signal. The serial was moved, as soon as the next time the `CondensedQueryExecutor` triggers a run it will start at the wrong point.

uweschaefer:
fixes #4127

### DIFF
--- a/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java
+++ b/factcast-store/src/main/java/org/factcast/store/internal/PgSynchronizedQuery.java
@@ -91,7 +91,7 @@ class PgSynchronizedQuery {
   }
 
   // the synchronized here is crucial!
-  @SuppressWarnings({"SameReturnValue", "java:51181"})
+  @SuppressWarnings({"SameReturnValue", "java:S1181"})
   public synchronized void run(boolean useIndex) {
     List<ConnectionModifier> filters =
         Lists.newArrayList(ConnectionModifier.withApplicationName(debugInfo));


bresciamattia (approved):
LGTM @uweschaefer, thanks

PR #4186, opened 2025-12-02
#4124: use single connection to fetch hwm and query facts

### LINKED ISSUE #4124: Improvement: call lastSerial always with dedicated SingleDataSource

We very often use the pattern:

1. get x as latest serial
2. run fact query
3. set factStreamPosition to x

in order to implement fast forwarding.

In order to be absolutely sure that the queries ran in 1. and 2. see the same data, we'd like to force them to use the same connection. This would need a refactoring of the LatestSerialFetcher and all the calls to it.

bresciamattia:
Closes #4124


factcast-core/src/test/java/org/factcast/core/subscription/observer/FastForwardTargetTest.java
mamup11: Shouldnt this class also be renamed?

PR #4208, opened 2025-12-15
#4151: transformationcache batched migration changesets

### LINKED ISSUE #4151: migrate_transformationcache_lastaccesses.sql needs to be replaced
**Describe the bug**
as we learned the hard way when rolling out.

We cannot block operation of the FactStore by running a 60sec update transaction, so we probably need to slice it down into chunks that run fast enough to be acceptable, and also not miss anything (as the TCache is a moving target when migrating online).

Hint: Probably using a tmp table for the IDs is necessary.

bresciamattia:
Closes #4151


uweschaefer:
@bresciamattia is there something left to do, or can this me mergend?

bresciamattia:
@uweschaefer feel free to review and merge, it is ready

Issue #4222, opened 2025-12-17
Simplify transformationcache access logging approach


uweschaefer:
Now, that we extracted the lastaccess data from the transformationcache into a sep. table, it might also be reasonable to simplify the apporach:

currently we do an "insert on conflict do nothing" and store/batch those on the eventstore side in order to batch those into a single transaction.

In order to get rid of this complexity, we could just do the update on retrieval if (once per day, only) this is necessary.
This could be easily encapsulated into the DB function. The performance impact is probably neglectable, but should be tested.

PR #4223, opened 2025-12-17
#4222: encapsulated last access tracking

uweschaefer:


factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -56,44 +55,42 @@ public class PgTransformationCache implements TransformationCache, AutoCloseable
...
-    this.bufferThreshold = bufferThreshold;
-    this.maxBufferSize = bufferThreshold;
     this.storeConfigurationProperties = storeConfigurationProperties;
+
+    registryMetrics.monitor(tpe, "transformation-cache");
+
+    this.maxBufferSize =
+        Math.max(maxBufferSize, 9999); // the batchUpdates used only support up to 10k
     this.buffer = new CacheBuffer(registryMetrics);
+    this.bufferThreshold = (THRESHOLD_PERCENT * maxBufferSize) / 100;
```
bresciamattia: I struggle to understand this, especially considering that the maxBufferSize parameter will always be 1000 (excluding tests), leading to additional calculations to just set this.maxBufferSize to 9999 and this.bufferThreshold to 800. Is there something I'm missing?

PR #4332, opened 2026-02-11
#4320: split catchup phase1

### LINKED ISSUE #4320: Split phase1 query to multiple queries, avoiding keeping cursors open for a long time
When doing a phase1 query, we basically run the query and keep the curor open until we sent all the event down to the client.
That means, for huge amounts of data or slow clients, this cursor might stay open for quite a while. This has bitten us before in the realm of tail-index-management.

We'd like to introduce another catchup strategy that runs the query only fetching serials, and the batch load the events for those serials and push them to the client.

This however might be problematic in terms of additional HEAP being used. An alterative variant of that is to query the serials to a temp table in the DB, so that the Heap is not going to be harmed. This unfortunately means, we'd need to stick to the same jdbc-connection, but at least the long running cursor is prevented and no additional (or potentially spiking) heap is needed.

uweschaefer:
closes #4320

### DIFF
--- a/factcast-store/src/main/java/org/factcast/store/internal/query/PgQueryBuilder.java
+++ b/factcast-store/src/main/java/org/factcast/store/internal/query/PgQueryBuilder.java
@@ -228,7 +228,7 @@ public class PgQueryBuilder {
 
   public String createSQL(long serial) {
 
-    if (tempTableName != null) {
+    if (useTemporaryTable()) {
       return "INSERT INTO "
           + tempTableName
           + "("
@@ -241,7 +241,8 @@ public class PgQueryBuilder {
           + createWhereClause()
           + AND
           + createSerialCriterionFor(serial);
-      // we don't need the order by here, because it will be ordered when reading
+      // we don't need the order by here, because it will be ordered when reading from the temp
+      // table
 
     } else
       return "SELECT "
@@ -255,6 +256,10 @@ public class PgQueryBuilder {
           + ORDER_BY
           + PgConstants.COLUMN_SER
           + " ASC";
+  }
+
+  private boolean useTemporaryTable() {
+    return tempTableName != null;
   }
 
   public String createStateSQL(long serial) {


factcast-store/src/test/java/org/factcast/store/internal/catchup/chunked/PgChunkedCatchupTest.java
```diff
@@ -0,0 +1,331 @@
...
+                  pipeline,
+                  serial,
+                  statementHolder,
+                  ds,
+                  PgCatchupFactory.Phase.PHASE_1));
+      doNothing().when(uut).fetch(any());
+
+      uut.run();
+
+      ArgumentCaptor<org.factcast.store.internal.pipeline.Signal> sigCap =
```
leflamm: qualification necessary?

factcast-store/src/test/java/org/factcast/store/internal/catchup/chunked/PgChunkedCatchupTest.java
```diff
@@ -0,0 +1,331 @@
...
+  SubscriptionRequestTO req;
+
+  @Mock @NonNull CurrentStatementHolder statementHolder;
+  @Mock @NonNull ServerPipeline pipeline;
+
+  @Mock(strictness = Mock.Strictness.LENIENT)
+  @NonNull
+  PgMetrics metrics;
+
+  @Mock @NonNull Counter counter;
```
leflamm: necessary?

factcast-site/documentation-docsy/content/en/Setup/properties.md
```diff
@@ -51,6 +51,7 @@ description: Properties you can use to configure FactCast
...
 | factcast.store.enumerationDirectModeEnabled                  | Despite of a Schema-Registry being defined or not, if set to true, enumeration of types or namespace will examine the data in the store directly, so that you only see data from already published facts.                                                                                                                                                                                                                                           | false                                    |
+| factcast.store.catchupStrategy                               | Available: FETCHING and CHUNKED. Fetching does the catchup query in one go and keeps the cursor open until the facts are sent to the client. Chunked runs queries limited to page-size rows instead.                                                                                                                                                                                                                                                | FETCHING                                 |
```
bedaka: I find the naming here not ideal because "fetching" and "chunked" don't sound like two alternatives for the same thing, since also chunks are fetched.
Maybe the counterpart of CHUNKED is better described as "straight" or sth similar

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunked/PgChunkedCatchup.java
```diff
@@ -0,0 +1,198 @@
...
+        .replace("$PROJECTION", PgConstants.PROJECTION_FACT)
+        .replace("$TMP", tempTableName)
+        .replace("$SIZE", Integer.toString(props.getPageSize()));
+  }
+
+  int prepareTemporaryTable(JdbcTemplate jdbc, String tempTableName) {
+    createTempTable(jdbc, tempTableName);
+
+    final var b = new PgQueryBuilder(req.specs(), statementHolder);
+    b.moveSerialsToTempTable(tempTableName);
```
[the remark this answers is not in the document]
uweschaefer: actually, it configured it to "moveSerialsToTempTable".

i renamed it.

PR #4345, opened 2026-02-17
Fix bug introduced in #4328

### LINKED ISSUE #4328: Experimental: prevent the generation of generic plans for catchup queries

From time to time, we see catchup queries to use index scans on the full GIN instead of using the tail indexes, even though a "ser>X" criterion is part of the query and should hint the planner to use the tails.

One possible lead of why PG would do In that 'occasionally' is described here:
https://www.rockdata.net/blog/custom-and-generic-plans/

In order to prevent the use of a generic plan, we'd like to try to make the "ser>X" criterion an un-parametrized part of the query in order to force PG to recreate the plan every time, taking the criterion with the concrete serial into account, thus realizing that the tails do in fact cover all possible tuples.

uweschaefer:
fixes #4328 where a query was rerun from the initial ser on, instead of the current.

### DIFF
--- a/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
+++ b/factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
@@ -1090,7 +1090,7 @@ class FactusClientTest extends AbstractFactCastIntegrationTest {
 
   /**
    * Testing, if a subscribed projection will receive duplicates. We check here if all events
-   * published are in fact consumed, and non of them pop up more than once.
+   * published are in fact consumed, and none of them pop up more than once.
    */
   @SneakyThrows
   @Test


factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
```diff
@@ -1087,4 +1087,42 @@ void noCallbackOnStarvation() {
       assertThat(callbacks).hasValue(0);
     }
   }
+
+  /** testing, if a subscribed projection will recieve duplicates. We check here if */
```
leflamm: recieve -> receive

factcast-itests/factcast-itests-factus/src/test/java/org/factcast/itests/factus/client/FactusClientTest.java
```diff
@@ -1087,4 +1087,42 @@ void noCallbackOnStarvation() {
...
+    factus.waitFor(subscribedProjection, new UUID(0, 99), Duration.ofSeconds(5));
+    assertThat(subscribedProjection.seen.size()).isEqualTo(100);
+  }
+
+  static class DuplicateChecker extends LocalSubscribedProjection {
+    final Set<UUID> seen = new HashSet<>();
+
+    @HandlerFor(ns = "test", type = "UserCreated")
+    synchronized void apply(Fact f) {
+      if (!seen.add(f.id())) throw new IllegalStateException("duplicate");
```
leflamm: would the test result be affected by an exception in all cases?

leflamm: please ignore, the test code waits explicitly for the final event to be successfully consumed

leflamm (approved):
LGTM. Thx!

PR #4538, opened 2026-04-21
#4362 Adding dynamic trace logging suppression

### LINKED ISSUE #4362: Make it possible to dynamically adjust the log level for "from scratch" catchups
In case of mass data being stored in the factstore, a query to rebuild a projection **from scratch** has to send a lot of data to the client. This for obvious reasons also creates a spike of emitted logs.
In cases where the amount of logs being processed in time is limited, people are tempted to adjust the log level from TRACE to DEBUG on the factcast server temporarily, when they know these rebuilds happen.
As this process needs manual intervention, is error prone and also a "one-for-all" solution in a sense, that factcast won't produce **any** trace logs during that time, we'd like to create another option.

If a dedicated property is set to a log level, we'd like to set an MDC attribute **in the thread doing a "from scratch" catchup**, so that a Filter configured into the appender can filter out all log events with a lower level than the configured if the MDC has the attribute mentioned above.

This way, we can filter only those logs, that belong to the "from scratch" process, while others work as expected.
If the property is unset, no MDC marking should be done.

mamup11:
closes #4362


factcast-store/src/main/java/org/factcast/store/registry/transformation/FactTransformerServiceImpl.java
```diff
@@ -95,58 +100,66 @@ public List<PgFact> transform(@NonNull List<TransformationRequest> req)
...
+                      pairs.parallelStream();
+
+                  try {
+
+                    // trying to avoid default FJP
+                    // https://blog.krecan.net/2014/03/18/how-to-specify-thread-pool-for-java-8-parallel-streams/
+
+                    return pool.submit(
+                            () ->
+                                pairStream
```
leflamm, line 170: genuinely-curious question 🙂 :

`callerMdc` seems to be set on the outer `CompletableFuture` worker, but the later `parallelStream()` work could run on other pool workers where that MDC was never set.

Or is it guaranteed that all parallel work keeps the caller MDC?

mamup11: That is true, I did it intentionally this way. Because the code inside the parallelStream doesnt have any logs, just in case of exceptions, so I thought about adding it just in case someone adds logs in the future. But decided against, since this context sharing between threads may be more of an issue than anything else.

Im happy to add it also to the inner logic, is just 2 lines to set and remove, but Im not sure if it makes sense

leflamm: understood

leflamm (approved):

Issue #4543, opened 2026-04-21
Add out-of-band flush in phase1 catchup to improve parallelization (client/server) at the cost of less eficient compression due to smaller messages


uweschaefer:
If the transfer buffer is big and the facts are very small, it may take many (10000) facts to fill up one message, that then will be flushed to the client.
This may lead to the client doing nothing waiting for the message, then being busy for some time (depending on the effort of processing).

If the messages were smaller in this case, it would help with the overall throughput as client and server work in parallel.

However, as this highly depends on the amount, and especially size of facts, as well as the question if they need to be transformed before sending, it is not easy to come up with the right chunk size for every situation.

The idea here is to insert extra flush signals into the stream after some time has passed, so that we don't make the situation worse where this was not applicable.

PR #4544, opened 2026-04-21
#4543: autoFlush added

uweschaefer:


factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
```diff
@@ -226,6 +227,12 @@ public enum CatchupStrategy {
 
   CatchupStrategy catchupStrategy = CatchupStrategy.CURSOR;
 
+  /** defines the max number of Facts being scheduled for transformation */
+  @Positive
+  @Max(60000)
+  @Min(AutoFlushingServerPipeline.AUTOFLUSH_CHECK_INTERVAL)
+  int autoFLushDelay = 10000; // 10 seconds default
```
otbe: FL?

factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
```diff
@@ -226,6 +227,12 @@ public enum CatchupStrategy {
 
   CatchupStrategy catchupStrategy = CatchupStrategy.CURSOR;
 
+  /** defines the max number of Facts being scheduled for transformation */
```
otbe: Is this comment correct?

otbe (commented):
Left some comments and dont we need an update of the docs in terms of properties?

factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
uweschaefer: totally intentional :thinking:

uweschaefer:
closes #4543

PR #4657, opened 2026-06-03
4589 Removing transformation cache purge job and ttl

### LINKED ISSUE #4589: Remove access tracking / TTL from transformation cache
Removal of transformation cache auto cleanup mechanisms 

Remove:

- The last-access timestamp table
- The join logic that reads from it
- The write-on-read path (updating last-access on every cache hit)
- The purge job and its configuration
- The old btree index on the full cache key (used by the invalidation trigger)

Add/change:

- created_at column on the transformation cache table, DEFAULT now(), never updated. 
- A fact_id column (UUID type) with a btree index on the transformation cache table. The existing DB trigger that deletes cached transformations when a fact gets invalidated stays, but it should use this new column instead of matching on the full cache key. The UUID index is smaller and faster to maintain.

**Additional context**

mamup11:
closes #4589


factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
```diff
@@ -0,0 +1,9 @@
+UPDATE transformationcache
+SET created_at = COALESCE(last_access, now());
+
+UPDATE transformationcache
```
uweschaefer: If we're lifting the fact_id, is there a way we also lift the version and path (as int array), so that we can have a composite key on those three instead of the string encoded one?
That'd give us the opportunity to use a BTREE again for lookups and  also delete by fact_id, while reducing the suspected size of the info.

Did not check the effect (in size) yet, though. Also it would require a two-step rollout to stay compatible, of course.

mamup11: I re used the test export I have for this tabled and created 4 tables to do a benchmark:

A) Current design (cache_key PK and HASH)
B) This branch change (cache_key PK, fact_id BTREE)
C) proposed design promoting more columns (fact_id, version, path, composite BTREE)
D) dropping version column since its not needed anyways (fact_id, path, composite BTREE)

Wins are marginal, basically 2-3% between option A and D. Basically what we save on size comes from the ditching the extra hash index 

table | heap | toast | indexes | total
-- | -- | -- | -- | --
V1 old (cache_key PK + HASH) | 119.4 MB | 17.3 MB | 14.4 MB | 151.6 MB
V2 branch (cache_key PK + fact_id) | 119.8 MB | 18.4 MB | 11.2 MB | 150.0 MB
V3 composite (fact_id, version, path) | 118.9 MB | 18.1 MB | 9.1 MB | 146.6 MB
V3b (fact_id, path) | 119.5 MB | 17.3 MB | 8.2 MB | 145.4 MB

Read path is even more marginal, but in general C was the slowest. In micro seconds:

batch | V1 old | V2 branch | V3 (fid,ver,path) | V3b (fid,path)
-- | -- | -- | -- | --
1 | 62 | 65 | 110 | 91
100 | 1468 | 1558 | 1715 | 1546
500 | 9205 | 7133 | 8997 | 7733

leflamm: Having a composite key seems cleaner to me, index will probably be smaller, invalidation by fact_id would be more efficient.

Two-step rollout concerns me a bit, would probably be:
1. Add nullable `fact_id`, `version`, `path` columns, still keeping `cache_key`
2. backfill from existing `cache_key`
3. write old and new columns
4. read new composite key during transition
5. enforce `NOT NULL`, add composite key unique constraint, switch the `ON CONFLICT` clause to the new compositkey
6. drop `cache_key` uniqness / idx

Can this be - if considered to be worth it - a follow-up ticket? Seems to bring lots of new complexity..

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
```diff
@@ -0,0 +1,9 @@
...
```
leflamm: Could it be possible that this unconditional update over 123 mio rows takes a little longer that expected?

We definitely had added "< 10m preconditions" to migrations that might be comparable in terms of effort to some degree (like 4123).

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
mamup11: I would argue that for the effort its not a change we should be including in this particular PR. But what is your opinion?

leflamm: I am surprised that the index size difference is not bigger between B and C... But since the difference it is rather small, I'd argue the potential gain of adding the composite index idea to this PR doesn't justify the added roll-out complexity. We can still argue to do it in a second step, though.

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
mamup11: Yep, this table is huge on prod, would probably take a lot of time, adding the pre condition

leflamm (approved):

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
uweschaefer: > Read path is even more marginal, but in general C was the slowest. In micro seconds:

Is this test pushed somewhere?

uweschaefer: checking with 73mio rows on a snapshot i see a somewhat different picture:


Schema:
```
postgres=> \d+ test_tc_comp2
                                        Table "public.test_tc_comp2"
 Column  |   Type    | Collation | Nullable | Default | Storage  | Compression | Stats target | Description 
---------+-----------+-----------+----------+---------+----------+-------------+--------------+-------------
 factid  | uuid      |           |          |         | plain    |             |              | 
 version | integer   |           |          |         | plain    |             |              | 
 path    | integer[] |           |          |         | extended |             |              | 
 header  | jsonb     |           |          |         | extended |             |              | 
 payload | jsonb     |           |          |         | external | lz4         |              | 
```




Table size is reduced from 160GB to 151GB

```code
postgres=> \dt+ transformationcache
                                           List of relations
 Schema |        Name         | Type  |   Owner   | Persistence | Access method |  Size  | Description 
--------+---------------------+-------+-----------+-------------+---------------+--------+-------------
 public | transformationcache | table | rdssystem | permanent   | heap          | 160 GB | 
(1 row)

postgres=> \dt+ test_tc_comp2 
                                        List of relations
 Schema |     Name      | Type  |   Owner   | Persistence | Access method |  Size  | Description 
--------+---------------+-------+-----------+-------------+---------------+--------+-------------
 public | test_tc_comp2 | table | rdssystem | permanent   | heap          | 151 GB | 

```

Index size is reduced from 7,2 to 5,5MB (or 2,8MB, if we try to get away without the path in the index)

```
postgres=> select * from stats_index 
where tablename like 'test_tc_comp2' or tablename like 'transformationcache';
 schemaname |      tablename      | num_rows | table_size |               index_name               | index_size | unique | num
ber_of_scans | tuples_read | tuples_fetched | valid 
------------+---------------------+----------+------------+----------------------------------------+------------+--------+----
-------------+-------------+----------------+-------
 public     | test_tc_comp2       | 73585320 | 101 GB     | idx_comp2_id_version                   | 2848 MB    | N      |    
           2 |           2 |              2 | Y
 public     | test_tc_comp2       | 73585320 | 101 GB     | idx_comp2_factid_version_path          | 5519 MB    | N      |    
           0 |           0 |              0 | Y
 public     | transformationcache | 73585320 | 33 GB      | transformationcache_pkey               | 5148 MB    | Y      |    
           0 |           0 |              0 | Y
 public     | transformationcache | 73585320 | 33 GB      | idx_transformationcache_cache_key_hash | 2048 MB    | N      |    
           0 |           0 |              0 | Y
(5 rows)
```

When it comes to the query execution:

```
postgres=> explain (analyze,buffers) select * from transformationcache where cache_key='093d2dcc-1479-4d1b-a8a6-db64e6122ae3-1-[3, 2, 1]';
                                                                          QUERY PLAN                                          
                                
------------------------------------------------------------------------------------------------------------------------------
--------------------------------
 Index Scan using idx_transformationcache_cache_key_hash on transformationcache  (cost=0.00..8.02 rows=1 width=817) (actual ti
me=0.009..0.010 rows=1 loops=1)
   Index Cond: ((cache_key)::text = '093d2dcc-1479-4d1b-a8a6-db64e6122ae3-1-[3, 2, 1]'::text)
   Buffers: shared hit=2
 Planning Time: 0.057 ms
 Execution Time: 0.025 ms
(5 rows)

```

vs

```
postgres=> explain (analyze,buffers) select * from test_tc_comp2 where factid='093d2dcc-1479-4d1b-a8a6-db64e6122ae3' and version=1 and path ='{3, 2, 1}';
                                                                   QUERY PLAN                                                 
                  
------------------------------------------------------------------------------------------------------------------------------
------------------
 Index Scan using idx_comp2_factid_version_path on test_tc_comp2  (cost=0.57..8.59 rows=1 width=1371) (actual time=0.014..0.01
4 rows=1 loops=1)
   Index Cond: ((factid = '093d2dcc-1479-4d1b-a8a6-db64e6122ae3'::uuid) AND (version = 1) AND (path = '{3,2,1}'::integer[]))
   Buffers: shared hit=5
 Planning Time: 0.080 ms
 Execution Time: 0.030 ms
(5 rows)

```

or 

```
postgres=> explain (analyze,buffers) select * from test_tc_comp2 where factid='093d2dcc-1479-4d1b-a8a6-db64e6122ae3' and version=1 and path ='{3, 2, 1}';
                                                              QUERY PLAN                                                      
         
------------------------------------------------------------------------------------------------------------------------------
---------
 Index Scan using idx_comp2_id_version on test_tc_comp2  (cost=0.57..8.59 rows=1 width=1371) (actual time=0.015..0.016 rows=1 
loops=1)
   Index Cond: ((factid = '093d2dcc-1479-4d1b-a8a6-db64e6122ae3'::uuid) AND (version = 1))
   Filter: (path = '{3,2,1}'::integer[])
   Buffers: shared hit=5
 Planning Time: 0.066 ms
 Execution Time: 0.030 ms
(6 rows)

```

Not much of a difference. The buffers are partly caused by different storage settings on header & payload (external instead of extended).

But of course there is no extra index needed for factid, as we're using btree.

```
postgres=> explain (analyze,buffers) select * from test_tc_comp2 where factid =  '093d2dcc-1479-4d1b-a8a6-db64e6122ae3';
                                                               QUERY PLAN                                                     
          
------------------------------------------------------------------------------------------------------------------------------
----------
 Index Scan using idx_comp2_id_version on test_tc_comp2  (cost=0.57..12.60 rows=2 width=1371) (actual time=0.012..0.014 rows=2
 loops=1)
   Index Cond: (factid = '093d2dcc-1479-4d1b-a8a6-db64e6122ae3'::uuid)
   Buffers: shared hit=6
 Planning Time: 0.055 ms
 Execution Time: 0.027 ms
(5 rows)
```

vs the (old and unfair comparison)

```
postgres=> explain (analyze,buffers) select * from transformationcache where cache_key like '093d2dcc-1479-4d1b-a8a6-db64e6122ae3%';
                                                                         QUERY PLAN                                           
                              
------------------------------------------------------------------------------------------------------------------------------
------------------------------
 Index Scan using transformationcache_pkey on transformationcache  (cost=0.69..1432747.05 rows=367927 width=817) (actual time=
0.016..0.017 rows=2 loops=1)
   Index Cond: (((cache_key)::text >= '093d2dcc-1479-4d1b-a8a6-db64e6122ae3'::text) AND ((cache_key)::text < '093d2dcc-1479-4d
1b-a8a6-db64e6122ae4'::text))
   Filter: ((cache_key)::text ~~ '093d2dcc-1479-4d1b-a8a6-db64e6122ae3%'::text)
   Buffers: shared hit=7
 Planning Time: 0.066 ms
 Execution Time: 0.034 ms

```

uweschaefer: Really don't want to make things more complex than necessary, but this PR as is in fact copies 160GB two times (alter table) and creates a new index (full scan) on a very busy prod table. if we're lifting all key elements instead of just one, it looks much cleaner to me as well as more efficient and to migrate, we'd just have to copy once.

Also, in order to ease the migration, we could remind ourselves that this is "only" cache data:

If we used an entire different table for the structural change, we could happily have two different instances running and using old and new approach, as well as we can roll back and forth nilly-willy.

Change my mind! :)

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
```diff
@@ -0,0 +1,9 @@
...
+SET created_at = COALESCE(last_access, now());
+
+UPDATE transformationcache
+SET fact_id = substring(cache_key, 1, 36)::uuid
+WHERE fact_id IS NULL;
+
+ALTER TABLE transformationcache
```
uweschaefer: There is a race condition in here, right? (Other instance can commit rows without fact_id)

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
uweschaefer: the updates are full table copies. it'd be better to combine them into one.

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
```diff
@@ -0,0 +1,9 @@
+UPDATE transformationcache
+SET created_at = COALESCE(last_access, now());
```
uweschaefer: this must not be executed within liquibase without precondition. On a snapshot, this update takes 28min being the only thing postgres does.

factcast-store/src/main/resources/db/changelog/factcast/issue4589/02_backfill.sql
mamup11: "Table size is reduced from 160GB to 151GB" This doesnt seem right to me, that doesnt come from the indexes changes alone, probably a different compression used? 
based on your result the indexes benefits are: 

```
 test_tc_comp2 | idx_comp2_id_version                   | 2848 MB 
 test_tc_comp2 | idx_comp2_factid_version_path          | 5519 MB
 transformationcache | transformationcache_pkey               | 5148 MB 
 transformationcache | idx_transformationcache_cache_key_hash | 2048 MB    

total | test_tc_comp2 8367MB | transformationcache 7196MB | delta: 1171MB

Thats closer to 13% at least more than 2-3% that I found on my test
```

mamup11:
Todos after sync:

Creation of a new table can run without issues
copy of data needs a pre condition in a transaction
Create indexes after migration, we decided btree on fact id and version. Index on the creation date column

code needs to be adapted the query and insert

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
```diff
@@ -34,7 +31,7 @@ public class InMemTransformationCache implements TransformationCache {
   // very low, but ok for tests
   private static final int DEFAULT_CAPACITY = 100;
 
-  private final Map<Key, FactAndAccessTime> cache;
+  private final Map<Key, PgFact> cache;
```
uweschaefer: could it be just Fact here instead of PgFact?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/TransformationCache.java
```diff
@@ -45,11 +42,14 @@ public interface TransformationCache {
...
 
-    String id;
+    UUID factId;
+
+    int version;
 
-    public static Key of(@NonNull UUID id, int version, @NonNull String transformationChainId) {
-      return new Key(
-          String.join("-", id.toString(), String.valueOf(version), transformationChainId));
+    // The transformation chain is deterministic for a given (factId, version), so it carries no
```
uweschaefer: i disagree. the path MUST be part of the key, as there are by definition many ways to get from A to B.

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
mamup11: Well, it could be swapped yes, but this class already depends on PgFact:

```
public Optional<PgFact> find(@NonNull TransformationCache.Key key) {
```

So I dont see a reason why the abstraction would be needed here. We could propagate the change to the interface and further. But then should that be done on this PR?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -154,12 +149,23 @@ public Set<PgFact> findAll(Collection<Key> keysToFind) {
...
+        sql.append(i == 0 ? "(?, ?, ?)" : ", (?, ?, ?)");
+      }
+      sql.append(")");
+
+      PreparedStatement ps = con.prepareStatement(sql.toString());
+      int idx = 1;
+      for (Key key : keys) {
+        ps.setObject(idx++, key.factId());
+        ps.setInt(idx++, key.version());
+        ps.setString(idx++, key.path());
```
uweschaefer: As discussed, i would see the path being an int array rather than a string

factcast-store/src/main/resources/db/changelog/factcast/issue4589/01_create_table.sql
```diff
@@ -0,0 +1,21 @@
+-- New transformation cache table keyed by (fact_id, version, path).
+-- Created empty so it never locks or copies the (potentially huge) old
+-- transformationcache table on startup. The old table is left untouched to
+-- keep rollback trivial.
+CREATE TABLE IF NOT EXISTS transformation_cache(
+    fact_id    uuid                     NOT NULL,
+    version    int                      NOT NULL,
+    path       text COLLATE "C"         NOT NULL,
```
uweschaefer: please make it an array of ints

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -275,7 +257,7 @@ void inTransactionWithLock(@NonNull Runnable o) {
         .execute(
             status -> {
               // we're using share mode here in order not to block reads from happening
-              jdbcTemplate.execute("LOCK TABLE transformationcache IN EXCLUSIVE MODE");
+              jdbcTemplate.execute("LOCK TABLE transformation_cache IN EXCLUSIVE MODE");
```
uweschaefer, line 262: it says shared in the comment but exclusive in the code. one of them is wrong :)

mamup11: Comment is wrong, changing

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -154,19 +149,30 @@ public Set<PgFact> findAll(Collection<Key> keysToFind) {
...
     return con -> {
-      PreparedStatement ps =
-          con.prepareStatement("select header, payload from selectTransformations( ? )");
-      Array idArray = con.createArrayOf("varchar", keys);
-      ps.setArray(1, idArray);
+      StringBuilder sql =
+          new StringBuilder(
+              "SELECT header, payload FROM transformation_cache WHERE (fact_id, version, path) IN (");
+      for (int i = 0; i < keys.size(); i++) {
+        sql.append(i == 0 ? "(?, ?, ?::int[])" : ", (?, ?, ?::int[])");
```
uweschaefer:
```suggestion
        if (i>0) sql.append(", ");
        sql.append("(?, ?, ?::int[])");
```

minor readability

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
```diff
@@ -219,9 +206,8 @@ public void invalidateTransformationFor(String ns, String type) {
       // it is fine if flush worked in another transaction, it just has to be serialized
       inTransactionWithLock(
           () ->
-              // will cascade down to tc_access
               jdbcTemplate.update(
-                  "DELETE FROM transformationcache WHERE header ->> 'ns' = ? AND header ->> 'type' = ?",
+                  "DELETE FROM transformation_cache WHERE header ->> 'ns' = ? AND header ->> 'type' = ?",
```
uweschaefer, line 211: Just wondering: isn't that going to take forever? (FTS)?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
uweschaefer: i think it is still wrong. if we grab exclusive, we lock out the shared ones.

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/PgTransformationCache.java
uweschaefer: will create a followup

Issue #3999, opened 2025-09-22
Enable offloading catchup PHASE1 to replica


uweschaefer:
**Is your feature request related to a problem? Please describe.**
When catching up, we use a two phase process. In P1 we look at the fact table until _now_ and return all those facts. In P2 we run the same query where ser> the last one we read in order to catch up with facts published during P1.

If P2 returns considerably less rows that are very likely to be buffered already anyway, we can argue that the majority of the load in the catchup phase comes from P1.

If we were able to direct those P1 queries to a read replica, we could profit in many ways:

* at least duplicate the capacity
* scale replica independently, maybe even dynamically
* improve with write throughput

At the same time, any problem that can be caused by replication lag is irrelevant, as we rerun the query on the primary in P2 anyway. More lag just means more rows from the primary.

PR #4666, opened 2026-06-04
#3999 "Cursor with Hold" chunking strategy

uweschaefer:


uweschaefer:
contributes to, but does not close #3999

uweschaefer:
see #4667

Issue #4667, opened 2026-06-04
add cursor with hold catchup strategy


uweschaefer:
As a prerequisite to #3999, we want to be able to use a chunked approach to catching up in phase 1 without hogging on the index for too long, but without using a temp table, as with prevents the use on a read-only (like read replica) instance of the DB, thus does not work for offloading.

using cursor with hold seems like a viable alternative.

PR #4668, opened 2026-06-04
#4667 Cursor with hold

### LINKED ISSUE #4557: Provide catchup strategy that - while not blocking the fact table for long - works without writing to postgres
We provided the CHUNKED Strategy next to the existing CURSOR Strategy, in order to be able to catch up (maybe from scratch) without blocking the fact table in a way that for instance index maint (create index concurrently) is no longer possible, because there was an open cursor.

While this works fine, it has two drawbacks:

* We use a temporary table to store IDs, which is considerable additional I/O, even in those cases, where there aren't many facts to fetch
* Using temp table we create and insert in has the effect that it does not work on a R/O instance (replica, maybe).

As a prerequisite to being able to offload PHASE1 to another instance, we need a strategy that is chuncked, but at the same time does not need to write to the database.

Implementation hint:

We don't want to pass serials back and forth from/to pgsql, so a CURSOR WITH HOLD looks like an attractive solution.

uweschaefer:


factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
```diff
@@ -212,6 +212,15 @@ public class StoreConfigurationProperties implements InitializingBean {
    */
   boolean readOnlyModeEnabled;
 
+  /**
+   * Size of a chunk, that is used to fetch events from the store during CHUNKED_HOLD catchup
```
leflamm:
```suggestion
   * Size of a chunk, that is used to fetch events from the store during CHUNKED_WITH_HOLD catchup
```

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -0,0 +1,320 @@
...
+      @NonNull PgCatchupFactory.Phase phase) {
+    super(props, metrics, req, pipeline, serial, statementHolder, ds, phase);
+    this.txMgr = txMgr;
+  }
+
+  @SneakyThrows
+  @Override
+  public void run() {
+    final var cursorName = createCursorName();
+    try (SingleConnectionDataSource sds =
```
leflamm: a very vaguely worried question 😆 

`declareCursor(...)` and `fetchChunk(...)` use that manually acquired connection, right?

Does `new TransactionTemplate(txMgr).execute(...)` use the same connection or another connection (because it is injected and backed by the "original" datasource)?

I seems worth a closer look to me because cursors are connection-local, aren't they?

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -0,0 +1,320 @@
...
+      @NonNull String fetchSQL,
+      @NonNull PgFactExtractor extractor,
+      @NonNull Timer.Sample timerSample,
+      @NonNull Timer timer,
+      @NonNull AtomicBoolean firstRowSeen)
+      throws SQLException {
+    int rows = 0;
+    try (Statement fetch = connection.createStatement()) {
+      fetch.setFetchSize(props.getPageSize());
+      statementHolder.statement(fetch, false);
```
leflamm: Unsure:

Should we run `statementHolder.statement(fetch, false)` with `rollbackOnCancel == false` in both different usages of `fetchChunk`? (first fetch vs. continued fetches) ? Value `false` definitely seems fine for the later fetches.

If anything bad happens in the TX that declares the cursor + fetches the first time would it be cleaner to rollback the TX and get rid of that cursor?

leflamm: different issue:
- `statementHolder.statement(fetch, false)` during the first use of `fetchChunk` might cause a commit because we treat the `wasCanceled()` case as normal
- could cause a cursor materialization even if we would not want that anymore

As discussed, trying to always use "true" might be a simple but valid approach

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -0,0 +1,334 @@
...
+        if (moreToFetch) {
+          log.trace("{} catchup {}, fetching further rows from held cursor", req, phase);
+
+          // thing is, we still need another transaction around, so that setFetchSize continues to
+          // work
+          //
+          // also we want to keep the transaction boundaries to one chunk only, in order not to
+          // block the
+          // index maintenance on the fact table
+          inTransaction(() -> continueFetchingUntilExhausted(cursor, extractor));
```
leflamm: I might well be wrong 🙂 but, it seems to me like 
- the while loop is inside the lambda passed to `inTransaction(...)` 
- `doInTransaction(...) ` starts one tx before the loop 
- the loop is what fetches all the remaining chunks
- only after the loop returns there is a commit

Do we need tx begins and commits happening inside each loop step to break down the tx boundaries into per-chunk as well?

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -0,0 +1,334 @@
...
+  private <R> R doInTransaction(@NonNull ThrowingCallable<R> callable) throws SQLException {
+    connection.setAutoCommit(false);
+
+    try {
+      R call = callable.call();
+      connection.commit();
+      return call;
+    } catch (SQLException sql) {
+      connection.rollback();
+      throw sql;
```
leflamm: Question: in this scenario

```java
  connection.setAutoCommit(false);
  callable.call();       // throws RuntimeException
  // commit is skipped
  // catch(SQLException) does not match exception type
  // rollback is skipped
  throw RuntimeException;
```

Would this mean that the tx remains "open" until some(?) cleanup eventually closes the connection?

factcast-store/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -595,13 +595,25 @@ databaseChangeLog:
...
+      changes:
+        - sqlFile:
+            encoding: utf8
+            path: factcast/issue4444/create_hash_idx_on_transformationcache.sql
+            relativeToChangelogFile: true
+            splitStatements: false
+            stripComments: true
+
+  - changeSet:
+      id: issue3999
```
leflamm: there were multiple branches (even issues?) involved, just in case we want to straighten this, this issue is `4557`

https://github.com/factcast/factcast/issues/4557

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
uweschaefer:
```
    } catch (Exception e) {
      connection.rollback();
      if (e instanceof SQLException sql) throw sql;
      throw new SQLException(e);
    }

```

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
uweschaefer: DUH! yes, that was exactly what the comment above suggested. Good catch, stupid me.

factcast-store/src/main/resources/db/changelog/db.changelog-master.yaml
uweschaefer: Still pondering if it makes sense to put this into a function in the schema, whereas the declaration is in java code. Looks not exactly symetric to my inner monk. What do you think @leflamm ?

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
leflamm: only noticed this because you told me about the plan before :smile:

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -0,0 +1,331 @@
...
+    if (connection == null)
+      throw new IllegalStateException("PgChunkedWithHoldCursorCatchup is a one-shot object");
+
+    try (Cursor cursor = createCursor(props.getChunkSize())) {
+      if (fetchAll(cursor)) {
+        log.trace("Done fetching, flushing.");
+        pipeline.process(Signal.flush());
+      }
+    } finally {
+      connection.close();
```
leflamm, line 70: very unsure about this, so a rather fuzzy question:

- if this `SingleConnectionDataSource` is configured with `suppressClose = true`
- and we close the connection here -> which maybe does not(?) close the underlying connection
- and we are on non-autocommit 

What happens to the cursor.close() in the try-with-resources in run():64

Asking myself if we should add a `inTransaction(cursor::close)` to the finally block before the `connection.close`...

factcast-store/src/main/resources/db/changelog/db.changelog-master.yaml
leflamm: I think i get your point: the cursor contract is now implicit.

Maybe a comment at cursor declaration a la “must match function contract: one bigint[] column per fetch” at least creates some sort of a connection.

To have more of a connection from the other side the function could also be named "fetchFactsFromChunkedHoldCursor"...?

factcast-store/src/main/resources/db/changelog/factcast/issue3999/fetchFactsFrom.sql
```diff
@@ -0,0 +1,28 @@
+-- The cursor cur is supposed to contain just one column of type bigint[] that
+-- contains all the serials of the facts to fetch in one function call.
+-- Those facts are then fetched and returned as a table, unless the cursor is exhausted.
+--
+-- See CHUNKED_WIH_HOLD Catchup Strategy
```
leflamm:
```suggestion
-- See CHUNKED_WITH_HOLD Catchup Strategy
```

leflamm (approved):

PR #4699, opened 2026-06-16
Issue3609: Introduce Blacklist refactoring via exclusion on database level

### LINKED ISSUE #3609: Filter blacklisted facts on database level
We're currently keeping an inmem list of event ids that were blacklisted in order to skip them. Turns out, that our blacklist grows really large. Even if this does not take too much heap within the JVM, there is extra work involved when reaading all those blacklisted events from the DB, just to skip them.

It'd be nice, to filter those on the query level already.

Hint: does not only involve fact streams, but also by-id lookups.

bedaka:
This change introduces the first step of migrating away from current mechanism called `blacklisting` that is based on a filter, towards a mechanism on database level (from now on referred to as `exclusion`) by:

- introducing the new column `exclusion_reason` in the fact table
- migrating the existing entries to the new column
- establishing a sync to keep both systems up to date
- introducing the `factcast.store.useInternalExclusion` property (default false) which 
    - switches off the old filter and
    - adds `exclusion_reason IS NULL` clause to fetching queries

Closes https://github.com/factcast/factcast/issues/3609


factcast-site/documentation-docsy/content/en/Concept/excluding-facts.md
```diff
@@ -1,26 +1,50 @@
...
+    );
+commit;
+```
+
+## Exclusion prior to version X.X
+
+If you use FactCast on a version prior to X.X, you have two ways for excluding (blacklisting) facts:
 
-## The postgres blacklist _(default)_
+### The postgres blacklist _(default)_
```
uweschaefer, line 47: we should mark as deprecated.

factcast-site/documentation-docsy/content/en/Concept/excluding-facts.md
```diff
@@ -29,7 +53,7 @@ Servers to make sure that changes take immediate effect.
 In order to document why the facts have been blacklisted, you can use the reason column (of type text). It will
 not be use for anything else, so there are no expectations on the content.
 
-## The filesystem blacklist
+### The filesystem blacklist
```
uweschaefer, line 56: we should mark as deprecated.

factcast-site/documentation-docsy/content/en/Concept/excluding-facts.md
```diff
@@ -1,26 +1,50 @@
...
+```
+
+## Exclusion prior to version X.X
+
+If you use FactCast on a version prior to X.X, you have two ways for excluding (blacklisting) facts:
 
-## The postgres blacklist _(default)_
+### The postgres blacklist _(default)_
 
 Blocked fact IDs can be added to a table named "blacklist" within the postgresDB. Inserting a new factId into the table
```
uweschaefer, line 49: shouldn't we deprecate this, or at least mention that from 13 on you set the reason instead?

factcast-site/documentation-docsy/content/en/Setup/properties.md
```diff
@@ -176,7 +177,9 @@ spring.grpc.server.keep-alive.permit-time=100
 
 ---
 
-### Blacklist
+### Blacklist (deprecated in version `0.X.X`)
```
uweschaefer: 0.13.0

factcast-site/documentation-docsy/content/en/About/migration.md
```diff
@@ -4,6 +4,26 @@ type = "docs"
 weight = 100015
 +++
 
+## Upgrading to 0.12.0
```
uweschaefer, line 7: 0.13.0

factcast-site/documentation-docsy/content/en/Setup/properties.md
```diff
@@ -70,6 +70,7 @@ description: Properties you can use to configure FactCast
...
 | factcast.store.offload.name                                         | same as any DataSource                                                                                                                                                                                                                                                                                                                                                                                                                              |                                          |
+| factcast.store.use-internal-exclusion                               | Enable only after finishing the blacklist migration described [here]({{< ref "migration.md">}}). Disables the blacklist entirely and instead enables filtering via exclusion column on the database level.                                                                                                                                                                                                                                          | false                                    |
```
uweschaefer, line 73: anchor missing

factcast-store/src/main/java/org/factcast/store/internal/filter/blacklist/Blacklist.java
```diff
@@ -38,8 +37,9 @@ public boolean isBlocked(@NonNull UUID factId) {
   @Override
   public void accept(@NonNull Set<UUID> currentList) {
     // we should not just replace it in order to not mess with the reference
-    Sets.SetView<UUID> toRemove = Sets.difference(blocked, currentList);
-    blocked.removeAll(toRemove);
-    blocked.addAll(currentList);
+    synchronized (blocked) {
+      blocked.retainAll(currentList);
```
uweschaefer, line 41: blocked.clear(); is probably easier & faster

factcast-store/src/main/java/org/factcast/store/internal/filter/blacklist/PgBlacklistDataProvider.java
```diff
@@ -31,24 +32,35 @@ public final class PgBlacklistDataProvider
...
   @Override
   public void afterSingletonsInstantiated() {
     bus.register(this);
-    updateBlacklist(); // initially necessary
+    if (!props.isUseInternalExclusion()) updateBlacklist(); // initially necessary
   }
 
   @Subscribe
   public void on(BlacklistChangeNotification signal) {
+    if (props.isUseInternalExclusion()) {
```
uweschaefer: should we maybe mitigate, by updating the reason in the fact table?

factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
```diff
@@ -216,6 +216,13 @@ public class StoreConfigurationProperties implements InitializingBean {
```
uweschaefer, line 216: i think, the properties set should be tested for plausibility:

using file based blacklist or jdbc blacklist vs internalExclusion (BlacklistConfigurationProperties)

uweschaefer: unless of course it is intended to use both. is it?

factcast-store/src/main/resources/db/changelog/factcast/issue3609/remove_blacklist.sql
```diff
@@ -0,0 +1,13 @@
+-- only run this once a rollback is out of question
```
uweschaefer, line 1: so this is meant to be added to liquibase in the next major?

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
uweschaefer, line 156: now that we might encounter excluded ones, this is not true anymore.

uweschaefer, line 159: even this is no longer correct, is it?

factcast-site/documentation-docsy/content/en/Setup/properties.md
uweschaefer: this is meant for the transition phase only, and will be removed afterwards, right?

uweschaefer (changes requested):

factcast-site/documentation-docsy/content/en/Setup/properties.md
bedaka: yes exactly

factcast-site/documentation-docsy/content/en/Concept/excluding-facts.md
bedaka: I expressed this by adding the headline `## Exclusion prior to version X.X` above
of course I have to update X.X to 0.13.0

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
bedaka: The queryBuilder used to create the Cursor in declare() is the one created in line 87 and contains the new property. this means the call to `queryBuilder.createSQL()`in line 219 contains the new condition and therefore no excluded facts are returned within fetchedRows.

So I think we do not have a problem

factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
bedaka: The approach I intended was to have this new migration property that will overwrite any blacklist config and instead enable the new mechanism. I did not add this into the existing blacklist config because

- StoreConfigProps are the properties that are already available when building the queryBuilder 
- the old blacklist config will be removed together with the new property eventually.

the alternatives would be to either introduce a third setting in the blacklist config and to wire it into all the places (I would prefer not to do that). Or to make it optional and ask people to set this new property to true and remove the blacklist config at the same time (this we could validate at startup).

factcast-store/src/main/resources/db/changelog/factcast/issue3609/remove_blacklist.sql
bedaka: yes the cleanup would prevent rollback without a migration so it is not intended to be applied automatically until the next breaking release.

factcast-site/documentation-docsy/content/en/Setup/properties.md
bedaka: added anchored link

PR #4716, opened 2026-06-25
factcast-spring-boot-autoconfigure: consider *all* BinaryJacksonSnapshotSerializerCustomizer beans for snapshot serialization

mkhl:
there's no apparent reason multiple customizers in the context need to be in competition
(except if one were to have several factcast/factus instances, but then one probably wouldn't use autoconfiguration anyway).

this change makes it so they aren't; instead it picks up *all* BinaryJacksonSnapshotSerializerCustomizer beans and combines them.


uweschaefer:
Welcome & thanks for the contribution. If you also create an issue for it, it'll trigger some internal workflows and is more likely to be picked up soon.

uweschaefer (approved):

uweschaefer:
see #4721

PR #4734, opened 2026-07-01
#4687 add async fetching

### LINKED ISSUE #4687: Add async fetching
# Context
When fetching lots of rows from the DB, we’re using the jdbc fetch-size to reduce the amount of communication with the database. when processing the rows, they are basically mapped to a PgFact and then passed along into the pipeline. Transformation etc, that might take some time to process, are already parallelized,.

However, the chunked approach seems fairly sensitive to the fetch size in terms of overall throughput, which made us experiment with huge fetch sizes to make up for this.

Obviously, huge fetchsizes also requires huge heap space and the larger the fetchsize gets, the harder it is to predict the maximum heap needed to be sure.

# The Goal
when having smaller fetch-sizes, we spend more time clocking the pipeline and talking to the database. As the processing is synchronous, we basically do (assuming fetch-size of 10 here)

```
map 1
map 2
... 
map 10
-> go fetching 11-20
map 11
map 12
...
```

Especially with CURSOR WITH HOLD the actual fetching might take some time, as it might involve extra queries.

What we want to do, is to fetch async to processing the rows, so that we minimize the time the processing is blocked by waiting for new data, while we still want to maintain the guarantee that fetch-size gives us (no more than $fetchSize rows buffered in resultset taking memory).

# Proposed solution
set fetch-size to page-size/2 and use a queue with pages, while fetching the next page already. Maybe RowSets can help.

Solution must prove to increase throughput
There are risks involved: added complexity, especially threading (exception handling). So in order for this to make sense, the throughput must significantly improve (at least 10% in laboratory conditions) for this to be accepted to main.

# Alternatives
We’re rejecting introducing r2dbc or similar solutions right away for now, as the risk involved in a change like that would be huge.

uweschaefer:
temporarily based on 4667


factcast-store/src/main/java/org/factcast/store/internal/catchup/tools/fetching/PreFetchingQuery.java
```diff
@@ -0,0 +1,156 @@
...
+    // same kind of heap memory, if we have one RS in processing
+    // and one waiting.
+    ps.setFetchSize(fetchSize / 2);
+
+    try (ps;
+        ResultSet resultSet = ps.executeQuery()) {
+      callbackBeforeProcessing.afterQueryFinished();
+
+      // async producer, terminated by closing the rs
+      CompletableFuture.runAsync(() -> produce(resultSet, q));
```
leflamm: Should we keep track of the returned `CompletableFuture` and cancel it when the consumer exits? I’m wondering whether the producer could otherwise remain blocked on `q.put()` if there’s nobody left to drain the queue.

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -246,50 +249,41 @@ SELECT array_agg(ser ORDER BY rn) FROM numbered GROUP BY grp ORDER BY grp ASC
...
+                    log.trace("{} catchup {}, fetch chunk statement was cancelled", req, phase);
+                  } else {
+                    PgFact fact = extractor.mapRow(rs, rows.get());
+                    pipeline.process(Signal.of(fact));
+                    rows.incrementAndGet();
+                  }
+                },
+                callbackAfterExecution::run);
+
+        return rows.get();
```
leflamm: Would an outer catch help with not escalating a cancellation, similar to the previous code?

```suggestion
        try {
          FetchingQuery.create(props)
              .executeAndProcess(
                  fetch,
                  rs -> {
                    if (statementHolder.wasCanceled()) {
                      log.trace("{} catchup {}, fetch chunk statement was cancelled", req, phase);
                    } else {
                      PgFact fact = extractor.mapRow(rs, rows.get());
                      pipeline.process(Signal.of(fact));
                      rows.incrementAndGet();
                    }
                  },
                  callbackAfterExecution::run);
        } catch (PSQLException e) {
          if (statementHolder.wasCanceled()) {
            log.trace("{} catchup {}, fetch chunk was cancelled", req, phase, e);
            return rows.get();
          }
          throw e;
        }

        return rows.get();

```

factcast-store/src/main/java/org/factcast/store/internal/catchup/tools/fetching/PreFetchingQuery.java
```diff
@@ -0,0 +1,156 @@
...
+    int fetchSize = ps.getFetchSize();
+    if (fetchSize == 0)
+      throw new IllegalArgumentException("Fetch size is not set on the PreparedStatement.");
+
+    ArrayBlockingQueue<ResultSet> q =
+        new ArrayBlockingQueue<>(1); // one in progress, one prefetched
+    // set to half the fetch size, so that we end up using the
+    // same kind of heap memory, if we have one RS in processing
+    // and one waiting.
+    ps.setFetchSize(fetchSize / 2);
```
leflamm: `StoreConfigurationProperties.pageSize` being 1 is a legal value AFAIU and `1 / 2 == 0` ... that might lead to trouble in line 152(?)

Would simply propose to use `ps.setFetchSize(Math.max(1, fetchSize / 2));`

factcast-store/src/main/java/org/factcast/store/internal/catchup/tools/fetching/PreFetchingQuery.java
uweschaefer: does not make much sense as this wont interrupt the producer. added it anyway and a comment about the mechanism of closing the RS before clearing the queue. this way, i think we're in the clear.

leflamm (approved):

PR #4746, opened 2026-07-06
#4745 use shared lock on unconditional publish

### LINKED ISSUE #4745: Share locks in unconditional publication
This will improve ingestion performance when publishing unconditionally, because currently the locking is done exclusively, which in fact should only be done in conditional publishing.

uweschaefer:
closes #4745

### DIFF
--- a/factcast-store/src/main/java/org/factcast/store/internal/lock/AdvisoryWriteLock.java
+++ b/factcast-store/src/main/java/org/factcast/store/internal/lock/AdvisoryWriteLock.java
@@ -26,6 +26,12 @@ public class AdvisoryWriteLock implements FactTableWriteLock {
 
   @Override
   @Transactional(propagation = Propagation.MANDATORY)
+  public void aquireSharedTXLock() {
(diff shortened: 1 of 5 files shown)


factcast-store/src/main/java/org/factcast/store/internal/lock/AdvisoryWriteLock.java
```diff
@@ -21,12 +21,24 @@
...
+  private static final String LOCK_SHARED_SQL =
+      "SELECT pg_advisory_xact_lock_shared(" + AdvisoryLocks.PUBLISH.code() + ")";
+  private static final String LOCK_EXCLUSIVE_SQL =
+      "SELECT pg_advisory_xact_lock(" + AdvisoryLocks.PUBLISH.code() + ")";
+
   private final JdbcTemplate tpl;
 
+  @Override
+  @Transactional(propagation = Propagation.MANDATORY)
+  public void aquireSharedTXLock() {
```
otbe: Im looking at the typo below since years and feel urged to fix it 😂

```suggestion
  public void acquireSharedTXLock() {
```

otbe (approved):

PR #4783, opened 2026-07-17
#4768 queued unconditional publish

### LINKED ISSUE #4768: Benchmark the use of server side batching of unconditional publications
The goal is to reduce the number of necessary db transactions in high concurrency publish scenarios. 

for each uncond. publish, we'd want to add a fact(s) to the query and then drain the queue in one batchInsert.
Lets see if the added complexity and execution path pays out.

uweschaefer:
closes #4768


factcast-store/src/main/java/org/factcast/store/internal/PgFactStore.java
```diff
@@ -114,54 +119,60 @@ public PgFactStore(
...
     }
 
+    if (props.getPublishBatch().isEnabled()) publishBatchable(factsToPublish);
+    else publishDirectly(factsToPublish);
+  }
+
+  void publishDirectly(@NonNull List<? extends Fact> factsToPublish) {
     metrics.time(
         StoreMetrics.OP.PUBLISH,
         () -> {
```
bedaka: fix sonar issue "useless curly braces"

factcast-site/documentation-docsy/content/en/Setup/properties.md
```diff
@@ -54,6 +54,8 @@ description: Properties you can use to configure FactCast
...
 | factcast.store.catchupStrategy                               | Available: CURSOR and CHUNKED_WITH_HOLD. Cursor does the catchup query in one go and keeps the cursor open until the facts are sent to the client. CHUNKED_WITH_HOLD first partitions serials into chunks and then fetched each chunk in a single query. Note that page-size still applies.                                                                                                                                                         | CURSOR                                   |
+| factcast.store.publishBatched.enabled                        | When enabled, the number of transactions can be reduced in burst publish situations and thus increase throughput, but adds a bit of latency to the publish operation                                                                                                                                                                                                                                                                                | false                                    |
```
bedaka:
```suggestion
| factcast.store.publishBatched.enabled                        | When enabled, the number of transactions can be reduced in burst publish situations. This increases throughput, but adds a bit of latency to the publish operation.                                                                                                                                                                                                                                                                                | false                                    |
```

factcast-store/src/main/java/org/factcast/store/internal/PgFactStore.java
```diff
@@ -390,4 +401,38 @@ public Long firstSerialAfter(@NonNull LocalDate date) {
       return null;
     }
   }
+
+  void batchPublish(List<? extends Fact> facts) {
```
bedaka, line 404: maybe mark protected (same for method below)

factcast-store/src/main/java/org/factcast/store/internal/PgFactStore.java
```diff
@@ -114,54 +119,60 @@ public PgFactStore(
...
+    else publishDirectly(factsToPublish);
+  }
+
+  void publishDirectly(@NonNull List<? extends Fact> factsToPublish) {
     metrics.time(
         StoreMetrics.OP.PUBLISH,
         () -> {
-          try {
-
-            List<Fact> copiedListOfFacts = Lists.newArrayList(factsToPublish);
```
bedaka, line 129: why no longer required?

factcast-store/src/main/java/org/factcast/store/internal/UnconditionalPublishQueue.java
```diff
@@ -0,0 +1,107 @@
...
+@Slf4j
+@RequiredArgsConstructor
+class UnconditionalPublishQueue {
+
+  private final PgFactStore pgFactStore;
+  private final int maxBatchSize;
+  private final ExecutorService flushingExecutor = Executors.newSingleThreadExecutor();
+  private final AtomicLong serialCounter = new AtomicLong(Long.MIN_VALUE);
+
+  record Publication(long serial, List<? extends Fact> facts, CompletableFuture<Void> completion) {}
```
bedaka: I might consider using a different term for `serial`  here as this is quite a loaded term in FactCast and we're not using this field for actual fact serials if I'm not mistaken

factcast-store/src/main/java/org/factcast/store/internal/UnconditionalPublishQueue.java
```diff
@@ -0,0 +1,107 @@
...
+      synchronized (queue) {
+        Publication p;
+        while ((pubs.size() < maxBatchSize) && (p = queue.poll()) != null) {
+          pubs.add(p);
+          facts.addAll(p.facts());
+        }
+      }
+
+      // could still be empty due to concurrent access
+      if (!pubs.isEmpty()) {
```
bedaka, line 95: Atm this is called from a singleThreadExecutor so concurrent flush() executions are not happening. 
Then flush() is not private so it is possible in theory that it might be called from somewhere else. Just wondered if we should update the comment or maybe make flush private & remove the guardrail or if it is there to be safe in case of future refactorings?

factcast-store/src/main/java/org/factcast/store/internal/UnconditionalPublishQueue.java
uweschaefer: right. ordinal?

factcast-store/src/main/java/org/factcast/store/internal/UnconditionalPublishQueue.java
uweschaefer: i'll change the comment

PR #4799, opened 2026-07-22
#3999 offloading of catchup phase1

### LINKED ISSUE #3999: Enable offloading catchup PHASE1 to replica

uweschaefer:


factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -207,32 +236,97 @@ void fastForward(@NonNull HighWaterMark atTheStartOfQuery) {
...
+      HighWaterMark highWaterMark = hwmFetcher.highWaterMark(connectionSupplier.dataSource());
+      // send FactStreamInfo if requested
+      if (request.streamInfo()) {
+        FactStreamInfo factStreamInfo = new FactStreamInfo(serial.get(), highWaterMark.targetSer());
+        pipeline.process(Signal.of(factStreamInfo));
+      }
+
+      if (!isConnected()) return;
+
+      // It is essential to provide SCSD to the catchup strategies.
```
otbe:
```suggestion
      // It is essential to provide SCDS to the catchup strategies.
```

factcast-store/src/main/java/org/factcast/store/StoreConfigurationProperties.java
```diff
@@ -281,6 +283,10 @@ public enum CatchupStrategy {
   @Max(50)
   long maxNotificationPollLatencyInMillis = 25;
 
+  public class OffloadDataSourceProperties extends DataSourceProperties {}
+
+  @Valid OffloadDataSourceProperties offload = new OffloadDataSourceProperties();
```
otbe, line 302: Should this be null or really with a default?

uweschaefer: default is cleaner i think. we can look for the URL directly and don't have to offload?,url.
Pattern becomes more obvious, if we have an "enabled" property.

factcast-core/src/main/java/org/factcast/core/subscription/observer/HighWaterMarkFetcher.java
```diff
@@ -16,25 +16,26 @@
...
 
 public interface HighWaterMarkFetcher {
   static HighWaterMarkFetcher forTest() {
-    return of(null, 0);
+    return forTest(null, 0);
   }
 
-  static HighWaterMarkFetcher of(UUID id, long ser) {
-    return Impl.of(HighWaterMark.of(id, ser));
+  static HighWaterMarkFetcher forTest(@Nullable UUID targetId, long targetSer) {
```
leflamm, line 29: Is it intended that the non-test class `FactStoreGrpcService` uses this method which is now called "forTest"?

factcast-store/src/test/resources/offload-datasource.properties
```diff
@@ -0,0 +1,5 @@
+# using the same as primay here
+factcast.store.offload.url=${spring.datasource.url}
```
leflamm, line 2: Do we want to keep these settings secret for now or should they already go into `factcast-site/documentation-docsy/content/en/Setup/properties.md`?

factcast-core/src/main/java/org/factcast/core/subscription/observer/HighWaterMarkFetcher.java
uweschaefer: it uses it in a @VisibleForTesting Constructor, so... kinda...

factcast-store/src/test/resources/offload-datasource.properties
uweschaefer: nope, just forgot. will add...

factcast-core/src/main/java/org/factcast/core/subscription/observer/HighWaterMarkFetcher.java
leflamm: I am blind...

leflamm (approved):

PR #4828, opened 2026-07-28
#4811: fix checks while closing

### LINKED ISSUE #4811: CurrentStatementHolder - statement is null, so no closing necessary. Duplicate call to close()?
Running PgFactStreamTelemetryTest for instance.
Check if this is due to half-arsed mocking, or if we actually have an asymetric set/close() somewhere.

bresciamattia:
Closes #4811


factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -15,81 +15,173 @@
...
+    Statement st = statement.get();
+    if (st != null) {
+      cancel();
+    }
+    // otherwise, there is nothing to do. Actually, it would be the expected state and behavior.
+  }
+
+  @VisibleForTesting
+  void clear() {
+    statement.set(null);
```
leflamm: Is is clear (harr) which statement `clear()` is clearing? This is called by `close()` from line 53.

Could it be, that "somebody" accidentally clears another statement? 

```java
    @Test
    void closingPreviousStatementDoesNotClearCurrentStatement() throws SQLException {
        var first = mock(PreparedStatement.class);
        var second = mock(PreparedStatement.class);

        var wrappedFirst = underTest.register(first);
        underTest.register(second);

        wrappedFirst.close();

        assertThat(underTest.statement()).isSameAs(second); // <-- this will fail
    }
```

I am not 100% sure this is an issue here but if feels kinda iffy.

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -15,81 +15,173 @@
...
+          // if we do not end the transaction, statements are canceled but still "idle in
+          // transaction" and so block further actions like wiping between tests
+          c.rollback();
+        }
       } catch (SQLException e) {
-        log.debug("While fetching connection from statement to cancel: {}", statement, e);
+        log.debug(
+            "Exception while rolling back transaction for cancelled statement {}:", statement, e);
+      } finally {
+        tryClose(c);
```
leflamm: This tries to close a connection from a data source originated in e.g. `PgSynchronizedQuery.java:105`

```java
    try (SingleConnectionDataSource ds = connectionSupplier.getPooledAsSingleDataSource(filters)) {
```

I am unsure whether it is safe to close the connection here? Is it even necessary?

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -15,81 +15,173 @@
...
-                "Exception while rolling back transaction for cancelled statement {}:",
-                statement,
-                e);
-          }
-      }
+  @VisibleForTesting
+  void cancelStatement(@NonNull Statement st) {
+    log.info("Canceling statement {}", st);
+    try {
+      st.cancel();
```
leflamm, line 121: Is it still possible to use `statementHolder.prepareStatement(connection, sql, queryBuilder.createStatementSetter(fromSerial))` here? If i am not mistaken something like...

```java
      var setter = queryBuilder.createStatementSetter(fromSerial);

      try (PreparedStatement declare = statementHolder.prepareStatement(connection, sql, setter)) {
        queryBuilder.createStatementSetter(fromSerial).setValues(declare);
        declare.execute();
        log.trace("{} catchup {}, cursor-with-hold declared", req, phase);
      }
```

... would restore the previous behavior where the declaration was registered before execution and could be cancelled when/if the subscription closes.

I think this was before indirectly achieved by 
Image
from `PgQueryBuilder.createStatementSetter()`

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
uweschaefer: The user of an instance of CSH is supposed to register and clear commands sequentially. Every PGStream has its own CSH, so that it is not prepared for "stacked" registrations.

If a second call to .register happens without clearing, we might think of adding a warning instead. Agree @leflamm ?

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
uweschaefer: yes, that is something i wanted to get rid of, because it is not an intuitive place for registering the statement. I'm gonna experiment with yet another different approach...

uweschaefer:
So, i changed a few things:

* Connection is wrapped and statements are tracked and unregister themselves on close.
* transaction is managed on close of the datasource
* PGSynchronizedQuery is no longer part of the mechanism, as it is expected to run very short-lived queries only, anyway.

I think this is somewhat cleaner now. However, during the analysis, i learned that PgConnection has the capability to **cancelQuery()** or **abort()**.

The first cancels the currently running query, but keeps the connection in a usable state. We could use that to drop the statement tracking altogether, but then we'd need to signal cancellation somehow for the Thread owning the statement not to try to continue.

abort() kills the connection entirely, so that we would not need any additional markers (every attempt to use the connection fails with "already closed" PSQLException being thrown), but also needs us to drop the connection from the pool entirely (discard in tomcat-jdbc-pool jargon). 
While that looks easier, it might introduce some "already closed" exceptions in the logs that raise eyebrows.

Not sure if any of those alternatives is much better, but i'll give it a try on a different branch, because it would reduce complexity and LOC quite a bit.

factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -309,7 +303,7 @@ void catchupPhaseTwo(PrimaryDataSourceSupplier primary, long phase1HighwaterMark
   @VisibleForTesting
   ModifiedSingleConnectionDataSource createCatchupDataSource(@NonNull DataSource ds) {
     return new ModifiedSingleConnectionDataSource(
-        ds.getConnection(), catchupConnectionModifiers(request));
+        statementHolder.track(ds.getConnection()), catchupConnectionModifiers(request));
```
leflamm, line 306: The tracked connection is aslo known to `ModifiedSingleConnectionDataSource` for `beforeReturn()` cleanup: https://github.com/factcast/factcast/blob/05de5e611f58aff0afec155cf0df9753a98119af/factcast-store/src/main/java/org/factcast/store/internal/listen/ModifiedSingleConnectionDataSource.java#L53

If `statementHolder.destroy()` runs during catchup, datasource destruction later tries to restore properties through this wrapper.

Questions
- might the cleanup statements fail with "already closed"?
- could this prevent `super.destroy()` on the next line then?

I am wondering if we need have the raw connection for cleanup stuff, like: `afterBorrow`, `beforeReturn`, potential rollbacks, super.destroy()... and have the tracked wrapper only for catchup queries?

leflamm: i first thought maybe a 

```java
  try {
    reversed.forEach(modifier -> modifier.beforeReturn(connection));
  } finally {
    super.destroy(); 
  }
```

would already solve this, but it only helps with returning the connection back to the pool. It would not reset existing modifiers or session properties. I do not know which values remain active on the next Postgres connection.

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
```diff
@@ -15,81 +15,123 @@
...
+    return new StatementTrackingConnection(connection, this);
+  }
+
+  // this is called destroy rather than close to make obvious that it can no longer be used at all
+  public void destroy() {
+    if (!wasDestroyed()) {
+      // if we still have a statement, we need to cancel it.
+      // note that cancelling a statement involves closing the connection, which is fine, as it
+      // happens only at the end of a PGStream usage.
+      Statement st = statement.get();
```
leflamm, line 60: probably very theoretical: could a new statement be registered between this line and line 67?

factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
uweschaefer: taken care of in the alternative approach

factcast-store/src/main/java/org/factcast/store/internal/query/CurrentStatementHolder.java
uweschaefer: obsolete in the alternative approach

PR #4859, opened 2026-08-08
#4811 alternative approach to remove statement tracking

### LINKED ISSUE #4811: CurrentStatementHolder - statement is null, so no closing necessary. Duplicate call to close()?

uweschaefer:
closes #4811


factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
```diff
@@ -274,21 +262,17 @@ int fetchChunk(@NonNull PgFactExtractor extractor, @Nonnull Runnable callbackAft
...
+                    rows.incrementAndGet();
                   },
                   callbackAfterExecution::run);
         } catch (SQLException e) {
-          if (statementHolder.wasCanceled()) {
-            log.trace("{} catchup {}, fetch chunk was cancelled", req, phase, e);
-            return rows.get();
-          }
-          throw e;
+          log.trace("{} catchup {}, fetch chunk encountered", req, phase, e);
```
leflamm: Could swallowing the SQLException (e.g. connection loss) here result in a partially fetched chunk being mixed-up with an exhausted cursor? 

E.g. if one fact is emitted and reading the next row fails, wouldn’t the returned partial row count cause the caller to flush the stream as complete?

P1 then returns its HWM and P2 starts after it, so...  wouldn’t the remaining facts be skipped permanently (rather than being recovered by the subsequent catch-up phase?)

factcast-store/src/main/java/org/factcast/store/internal/pipeline/PushbackServerPipeline.java
```diff
@@ -0,0 +1,58 @@
...
+  public void close() {
+    isClosed.set(true);
+    delegate.close();
+    // we don't close the datasource, because it will be done in the catchup process.
+    onCloseListeners.forEach(CatchupDataSource::cancel);
+  }
+
+  // these two methods are use to communicate a close to a datasource in use. Note that due to
+  // offloading, there might be two separate DS valid at a time.
+  public void register(@NonNull CatchupDataSource ds) {
```
leflamm: Could a pipeline be closed before the catch-up data source is registered?

Asking because if so, `close()` would find no datasources to cancel, datasource then gets registered, register does not notice that the pipeline has been closed and the datasource/query are never cancelled...

```java
  @Test
  void cancelsDataSourceRegisteredAfterPipelineWasClosed() {
    CatchupDataSource dataSource = mock(CatchupDataSource.class);
  
    underTest.close();
    underTest.register(dataSource);
  
    verify(dataSource).cancel();
  }
```

factcast-store/src/test/java/org/factcast/store/internal/QueryCancellationIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+            @NonNull ServerPipeline pipeline,
+            @NonNull AtomicLong serial,
+            @NonNull SingleConnectionDataSource ds,
+            @NonNull Phase phase) {
+          return new PgCursorCatchup(props, metrics, request, pipeline, serial, ds, phase) {
+
+            @Override
+            protected PgQueryBuilder createPgQueryBuilder(List<FactSpec> specs) {
+              return new PgQueryBuilder(specs) {
+                public String createSql() {
```
leflamm: method appears to be unused... is this supposed to be
```suggestion
                @Override
                public String createSQL() {
```
?

factcast-client-grpc/src/main/java/org/factcast/client/grpc/ResilientGrpcSubscription.java
```diff
@@ -203,7 +203,10 @@ void fail(Throwable exception) {
     close();
     currentSubscription.unblock();
     originalObserver.onError(exception);
-    throw ExceptionHelper.toRuntime(exception);
+
+    // we do not rethrow the exception here, because the resources have been cleaned up, the
```
leflamm: I understand(?) this reasoning is focussed around asynchronous `DelegatingFactObserver.onError()`, like: a subscription is connected and receives facts and a (non-retryable) error happens...

Is it intended that this also changes **synchronous** startup failures 
- from: `subscribe()` throws
- to: `subscribe()` returns a closed subscription and only calls `observer.onError()`

?

With "synchronous" I mean something like
```
  GrpcFactStore.subscribe()
    -> new ResilientGrpcSubscription()
      -> connect()
        -> initializeAndConnect()
          -> store.initializeIfNecessary()
```

factcast-store/src/main/java/org/factcast/store/internal/pipeline/PushbackServerPipeline.java
uweschaefer: interesting. Dunno how exactly, but certainly a good way of bullet proofing

factcast-store/src/test/java/org/factcast/store/internal/QueryCancellationIntegrationTest.java
uweschaefer: DUH

factcast-client-grpc/src/main/java/org/factcast/client/grpc/ResilientGrpcSubscription.java
uweschaefer: in my book it would make sense, because the application code needs to react on onError anyway. But you're right, maybe that is too courageous to also add here. I'll create a followup issue for this change.

factcast-store/src/main/java/org/factcast/store/internal/catchup/chunkedwithhold/PgChunkedWithHoldCursorCatchup.java
uweschaefer: true, i think the whole "how to end this gracefully" needs a bit more polishing.

uweschaefer: actually, p2 will start wherever p1 stopped. no ffwd in between. If however both would be cancelled, we're in a bad place.

factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -324,18 +351,15 @@ long catchupPhaseOne(@NonNull SingleConnectionDataSource dataSourceToUseForP1) {
...
+          .create(request, pipeline, serial, dataSourceToUseForP1, PgCatchupFactory.Phase.PHASE_1)
           .run();
 
       // serial might be higher than hwm, because of concurrent inserts, but it also may be much
-      // smaller
-      // in which case, we want to continue from hwm, to not do unnecessary work
+      // smaller in which case, we want to continue from hwm, to not do unnecessary work.
+      //
+      // Note that any kind of exceptional behavior like cancellation, random SQLExceptions or the
+      // like are expect to THROW, so that a "between phases ffwd" only happens, if we know that
```
leflamm, line 361: > `// are expect to THROW...`

Is this true for `PgChunkedWithHoldCursorCatchup`? `fetchChunk()` (starting at line 256) still catches the `SQLException` and returns the rows fetched so far. Wouldn’t `catchupPhaseOne()` then return the HWM even though the fetch failed, which is what this comment says must not happen?

Or am I mixing stuff up here?

factcast-store/src/main/java/org/factcast/store/internal/pipeline/PushbackServerPipeline.java
```diff
@@ -0,0 +1,63 @@
...
+
+  // these two methods are use to communicate a close to a datasource in use. Note that due to
+  // offloading, there might be two separate DS valid at a time.
+  public void register(@NonNull CatchupDataSource ds) throws PipelineAlreadyClosedException {
+
+    if (isClosed.get()) throw new PipelineAlreadyClosedException();
+
+    if (!onCloseListeners.add(ds))
+      throw new IllegalStateException("DS was already registered. This is a weird bug.");
+  }
```
leflamm: could `close()` happen between checking `isClosed` and adding the listener? wouldn’t we then add the datasource after `close()` already processed the empty listener set, so it never gets cancelled?

maybe we add it first and check `isClosed` afterwards, cancelling it right away if necessary? `cancel()` would then probably need to be idempotent...


```suggestion
  public void register(@NonNull CatchupDataSource ds) {
    if (!onCloseListeners.add(ds))
      throw new IllegalStateException("DS was already registered. This is a weird bug.");
      
      if (isClosed.get()) {
        ds.cancel();
      }
  }
```

factcast-store/src/main/java/org/factcast/store/internal/catchup/CatchupDataSource.java
```diff
@@ -0,0 +1,161 @@
...
+        log.warn("Unwrapping of PgConnection failed. This is ok, if we're in a unit test");
+      else pgNative.cancelQuery();
+    } catch (SQLException e) {
+      log.warn("Cancelling of orphaned query failed on datasource destruction ", e);
+    }
+  }
+
+  /** This may be called from another thread. */
+  public void cancel() {
+    this.wasCanceled.set(true);
```
leflamm, line 151: in case we want cancel() to be idempotent... see discussion here https://github.com/factcast/factcast/pull/4859/changes#r3881546746

```suggestion
  public void cancel() {
    if (!wasCanceled.compareAndSet(false, true)) {
      return;
    }
```

factcast-store/src/main/java/org/factcast/store/internal/pipeline/PushbackServerPipeline.java
uweschaefer: yep that would be possible double cancellation. actually this is all supposed to happen around one thread. as "contentionless synchronization is virtually free", i'll just add synchronized to reduce headaches.

factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
uweschaefer: crap, work half done - i meant to remove the catching business once and for all

factcast-store/src/main/java/org/factcast/store/internal/pipeline/PushbackServerPipeline.java
```diff
@@ -0,0 +1,62 @@
...
+  private boolean isClosed = false;
+  private Set<CatchupDataSource> onCloseListeners = Collections.synchronizedSet(new HashSet<>());
+
+  public PushbackServerPipeline(@Nonnull ServerPipeline chain) {
+    this.delegate = chain;
+  }
+
+  public synchronized void process(@NonNull Signal s) throws PipelineAlreadyClosedException {
+    if (isClosed) throw new PipelineAlreadyClosedException();
+    delegate.process(s);
```
leflamm, line 34: do you  think it is worth considerung leaving the `delegate.process(s);` outside the synchronization?

in case this would take a long time (waiting for whatever... transformation, client, network), could it delay closing?

uweschaefer: Right. It could delay closing.

leflamm: I withdraw this!

I think the synchronization is fine here, normal cancellation should unblock the processing quickly, and moving `delegate.process()` outside would of course not actually stop something like a stuck transformation anyway

factcast-store/src/main/java/org/factcast/store/internal/catchup/CatchupDataSource.java
```diff
@@ -0,0 +1,161 @@
...
+  public CatchupDataSource(
+      @NonNull Connection connection,
+      @NonNull List<ConnectionModifier> modifiers,
+      @NonNull PushbackServerPipeline pipeline) {
+    super(connection, modifiers);
+    this.connection = connection;
+    this.pipeline = pipeline;
+
+    // we want to know when the pipeline is closed, so that we can cancel and abort.
+    pipeline.register(this);
```
leflamm, line 64: could this leak the connection if `register(...)` throws because the pipeline was closed in between? 

the connection was already obtained from the pool and modified, but as the constructor fails, the caller never receives the datasource and cannot close it.

should we catch the RuntimeException here, call `super.destroy()` and rethrow it?

maybe somethink like

```suggestion
      try {
    pipeline.register(this);
  } catch (RuntimeException registrationFailure) {
    try {
      super.destroy();
    } catch (RuntimeException cleanupFailure) {
      registrationFailure.addSuppressed(cleanupFailure);
    }
    throw registrationFailure;
  }

```

leflamm (approved):

leflamm:
LGTM! Thx @uweschaefer !

PR #4864, opened 2026-08-10
#3882: issue3882: introduce revisionId string as an alternative to re…

### LINKED ISSUE #3882: Add String field for revision to ProjectionMetaData
In case you want to provide aggregates or projections as dependencies, you might end up in a situation where you need to provide a fix for an earlier version, and then it gets complicated to find the next free revision for ProjectionMetadata, as on other branches, it might already be in use. 

You litereally need to scan all branches for that, every time.

Add a field `String revisionStr` to ProjectionMetadata (I'm open for better names :D)

For backward-compatibility, we must choose one of the two approaches:

Either, allow that you can set both fields, since `revision` is 0 if not set explicitly, and include both if `revisionStr` is set (otherwise just `revision`, with default 0)

Or, allow those three cases:
- nothing set (then use `revision`with default 0)
- only `revision` set explicitly
- only `revisionStr` set explicitly

Then either use `revisionStr` if set explicitly and omit `revision` (which has value 0), or use only `revision`.

The field is only used in ScopedName so far, so could be easily added (but should be communicated as breaking change, if somebody made use of it outside of factcast). Then the change in ScopedName for the first approach would just be (omitting the checks mentioned above):

```java
String optionalRevision = Optional.ofNullable(metaData.revisionStr())
    .map(r -> NAME_SEPARATOR + r)
    .orElse("");

return ScopedName.of(name + NAME_SEPARATOR + metaData.revision() + optionalRevision);
```
For the second approach, use either `revision` or `revisionStr`.

Like this, should be backward-compatible.

bedaka:
…vision.


factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
```diff
@@ -27,15 +27,33 @@
...
 @SuppressWarnings("java:S1133")
 public @interface ProjectionMetaData {
+
+  String DEFAULT_REVISION_ID = "0";
+
   String name() default "";
 
-  /** will be made required as soon as serial() is removed */
+  /** will be removed in favor of revisionId in the future */
+  @Deprecated
```
mamup11, line 36: I like the idea of supporting both, and both being exclusive. But why do we want to deprecate this and leave only the string one?

factcast-factus/src/main/java/org/factcast/factus/projection/ScopedName.java
```diff
@@ -44,7 +44,16 @@ public static ScopedName fromProjectionMetaData(Class<?> clazz) {
       name = ClassUtils.getNameFor(clazz);
     }
 
-    return ScopedName.of(name + NAME_SEPARATOR + metaData.revision());
+    return ScopedName.of(name + NAME_SEPARATOR + revisionIdentifier(metaData));
+  }
+
+  private static @NonNull String revisionIdentifier(@NonNull ProjectionMetaData metaData) {
+    // 0 is the default value therefore we assume revision is not set.
+    if (metaData.revision() != 0) {
```
mamup11, line 52: >= 0 to account for negatives as well?

factcast-factus/src/test/java/org/factcast/factus/projection/ScopedNameTest.java
```diff
@@ -64,11 +68,31 @@ void testWitherRefusesEmpty() {
         .isInstanceOf(IllegalArgumentException.class);
   }
 
-  @ProjectionMetaData(revision = 2)
```
mamup11, line 67: You remove the tests of revision as well? shouldnt we keep both?

factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
```diff
@@ -27,15 +27,33 @@
...
-      return Optional.ofNullable(clazz.getAnnotation(ProjectionMetaData.class));
+      final Optional<ProjectionMetaData> metaData =
+          Optional.ofNullable(clazz.getAnnotation(ProjectionMetaData.class));
+      metaData.ifPresent(md -> validate(clazz, md));
+      return metaData;
+    }
+
+    private static void validate(Class<?> clazz, ProjectionMetaData md) {
+      boolean hasRevisionSet = md.revision() > 0;
+      boolean hasIdSet = !md.revisionId().equals(DEFAULT_REVISION_ID);
```
mamup11, line 52: Probably should also validate empty here, currently passing revisionId = "" will result as hasIdSet true

factcast-site/documentation-docsy/content/en/About/migration.md
```diff
@@ -4,6 +4,62 @@ type = "docs"
...
+@ProjectionMetaData(revisionId = "issue1234")
+public class MyProjection extends AbstractManagedProjection { ... }
+```
+
+`revision` and `revisionId` are **mutually exclusive** and setting both will resolve in an `IllegalArgumentException`
+when the projection is first resolved.
+
+{{% alert title="Keep the value identical" theme="warning" %}}
+The revision is part of the key under which projection state and snapshots are persisted. `revisionId = "1"` produces
+exactly the same scoped name as `revision = 1` did, so migratin is stable. If you decide to change the name during
```
mamup11, line 31:
```suggestion
exactly the same scoped name as `revision = 1` did, so migration is stable. If you decide to change the name during
```

factcast-factus-migration/src/test/java/org/factcast/factus/migration/RevisionToRevisionIdRecipeTest.java
```diff
@@ -0,0 +1,331 @@
...
+            """
+            import org.factcast.factus.serializer.ProjectionMetaData;
+
+            @ProjectionMetaData(revision = -1)
+            class MyProjection {}
+            """,
+            """
+            import org.factcast.factus.serializer.ProjectionMetaData;
+
+            @ProjectionMetaData(revisionId = "-1")
```
mamup11, line 240: Im confused here. In the description of the recipe, it says "Only non-negative integer literals are rewritten" but in the migration.md I also didnt saw negative numbers explicitly excluded.

factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
bedaka: I think there is not really a benefit in supporting two fields that essentially do the same just in two different types.

factcast-factus/src/main/java/org/factcast/factus/projection/ScopedName.java
bedaka: Yes, 0 is the default value, anything else would be a user provided value, which we want to keep in order to not change the current scoped name

factcast-factus/src/test/java/org/factcast/factus/projection/ScopedNameTest.java
bedaka: good point

factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
bedaka: added check

factcast-factus-migration/src/test/java/org/factcast/factus/migration/RevisionToRevisionIdRecipeTest.java
bedaka: I changed this during implementation and forgot to update the description. thanks for pointing it out

factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
mamup11: Hmmm, I grew used to the numeric value as a safe value to increase. So I thought keeping both to keep my numeric value is worth it. Its true a string value does the same job, perhaps is ok to just keep that one

mamup11 (approved):

factcast-factus/src/main/java/org/factcast/factus/serializer/ProjectionMetaData.java
bedaka: Yes the point of this change is to actually increase this safety by not only changing 1 -> 2 but to be able to add some unique suffix like "2-issue1234" this way we remove the risk of collision with a second parallel commit that also moves to 2 but with a completely different changeset

mamup11 (changes requested):
Ah, didnt noticed the sonnar complains

bedaka:
@mamup11 the sonar complaints cannot be fixed without removing the `@Deprecation` reason that sonar asked for beforehand. I added `@SuppressWarning`

mamup11 (approved):

PR #4870, opened 2026-08-11
#4865: honor factcast.store.offload.enabled for catchups

### LINKED ISSUE #4865: factcast.store.offload.enabled is documented but not honored
**Describe the bug**

The configuration property `factcast.store.offload.enabled` [is documented with a default value of false](https://github.com/factcast/factcast/blob/0.12.0/factcast-site/documentation-docsy/content/en/Setup/properties.md#L64), but it is not evaluated by the implementation.

Offloading is currently activated solely by the presence of `factcast.store.offload.url`. So, configuring a URL together with `factcast.store.offload.enabled=false` does enable phase-1 catchup offloading.

**Expected behavior**
  - `factcast.store.offload.enabled=false` disables offloading, even when the datasource properties are configured.
  - `factcast.store.offload.enabled=true` enables offloading and requires a valid offload datasource configuration.

leflamm:


uweschaefer (approved):
lMBtm, thx

PR #4913, opened 2026-09-03
#4912: handle exceptions in notification timer tasks

### LINKED ISSUE #4912: Prevent database failures from killing the notification polling timer
**Describe the bug**
When the database connection fails during a scheduled notification poll, the exception is not handled and kills the timer thread. All following nudges fail with Timer already cancelled, and polling only recovers after a restart.

0.12.1

**Expected behavior**
A temporary database connection failure should only fail the current poll. The timer should remain active and continue processing following nudges.

**Additional context**

Image

Image

Image

leflamm:


uweschaefer (approved):

uweschaefer:
thx @leflamm

PR #4937, opened 2026-09-08
issue4883: fix flaky pipeline behaviour

### LINKED ISSUE #4883: Investigate flaky test not finding an artifact
https://github.com/factcast/factcast/actions/runs/31714519609/job/94495700651?pr=4882

bedaka:
The origin of the issue is rooted in the combination of the `postgres-compatibility`  not executing a `install` but relying entirely on the cache in combination with the `factcast-schema-registry-cli` pom.xml defining `spring-boot-thin-layout`.

- When running the integration-test the thin jar requires its dependencies to be present in the ~/.m2, which entirely depends on the cache that was pulled by the job. 
- Since the `build` job (which executes `mvn install`) and `postgres-compatibility` run in parallel it will always restore the cache from a previous run - if this run does not contain the required artifacts it will fail (e.g. after a version bump).

Since the additionall install step probably was skipped to prevent the additional work I opted for executing the "flaky tests" based on the reactor classpath to be independent from the pulled cache.

Closes #4883

### DIFF
--- a/factcast-schema-registry-cli/pom.xml
+++ b/factcast-schema-registry-cli/pom.xml
@@ -323,9 +323,15 @@
             <phase>integration-test</phase>
             <configuration>
               <executable>java</executable>
+              <!-- Runs off the reactor classpath instead of within repackaged jar,
+                since that would make the thin launcher resolve this module's parent POM
+                from the local/remote repository at runtime, which fails in any build that
+                did not "install" first (e.g. a bare "mvn verify"). -->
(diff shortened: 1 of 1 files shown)


bedaka:
What we might loose with this solution is that the packaged jar is not tested directly - but I figured this is a more reasonable tradeoff than adding a dedicated install step to the integration pipeline just for this module.

uweschaefer:
thx @bedaka

PR #4942, opened 2026-09-09
propagate follow query failures 

leflamm:
If a follow query fails to get a database connection, `QueryExecutor.runTarget()` logs the error, but does not tell the subscriber.

The GRPC client has retry logic, but without an error it has no reason to reconnect. Keepalives can still arrive, so the subscription looks alive while it does not receive new events.

Another matching notification might get it going again. if none arrives, it can stay stuck.

We might need to let the target fail and propagate the error so the client can react.

The added IT fails one connection attempt and waits for "onError". Without the proposed fix, that never arrives.


uweschaefer:
we decided to not propagate.

PR #4946, opened 2026-09-11
Dont fail on missing credentials

### LINKED ISSUE #4945: Do not throw exception if a secret is not available for a role

otbe:
closes #4945

### DIFF
--- a/factcast-server-security/src/main/java/org/factcast/server/security/CommonSecurityConfiguration.java
+++ b/factcast-server-security/src/main/java/org/factcast/server/security/CommonSecurityConfiguration.java
@@ -98,7 +98,7 @@ public class CommonSecurityConfiguration {
       return account
           .flatMap(
               a ->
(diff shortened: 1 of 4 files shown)


factcast-server-security/src/main/java/org/factcast/server/security/CommonSecurityConfiguration.java
```diff
@@ -148,16 +148,15 @@ private static FactCastAccessConfiguration parseAccessConfiguration(
       List<String> ids = cfg.accounts().stream().map(FactCastAccount::id).toList();
       for (String id : ids) {
         if (!accessSecrets.getSecrets().containsKey(id)) {
-          throw new IllegalArgumentException("Missing secret for account: '" + id + "'");
+          log.warn("Missing secret for account: '{}'", id);
```
mamup11, line 151: I will call this an error log, so at least we get notified

mamup11 (approved):

PR #4966, opened 2026-09-15
#4872: more selective purging of stale versions

### LINKED ISSUE #4872: Performance/Load bug: deleting from transformationcache on either manual manipulation of the transformationcache or by replacing a schema
Currently, tr_deferred_transformationstore_change is triggered on update/deletion on transformationstore. in such a case, we'd need to delete all transformation results that are/ or include this particular version.

As before we did not have this information in the transformationcache filterable (but only as a string-encoded key) we preferred to 
```java
 inTransactionWithLock(
          () ->
              jdbcTemplate.update(
                  "DELETE FROM transformation_cache WHERE header ->> 'ns' = ? AND header ->> 'type' = ?",
                  ns,
                  type));
```

This is terrible in 3 regards:

1. It is a full table scan, as none of that information is indexed
2. we delete too much (completely wipe (ns,type) where ns,type and all rows that CONTAIN the given version in their path would be enough)
3. During that time, the TC table is exclusively locked

Suggestion to fix this:

1. add an index on ns/type to limit the scan.
2. extend the tr_deferred_transformationstore_change to send also the version (includes TransformationStoreChangeNotification) and limit the deletion to just those rows, that have the version in their transformation path.
3. create a procedure to call from java, which does a select first, and after acquires the lock and deletes. This way, the lock span should be minimized.

Hint: having a ns and type columns is probably wasteful as we need it just for this usecase. we can just direct the index to random projected attributes:
```sql
CREATE INDEX CONCURRENTLY idx_transformation_cache_ns_type_path
ON transformation_cache (
    (header ->> 'ns'),
    (header ->> 'type')
);
```
Adding the path to the index would only help if we use a GIN as it supports conatins lookups, but the maintenance cost of that would be huge, so that indexing btree on ns/type is probably favorable.

uweschaefer:
closes #4872


factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
```diff
@@ -0,0 +1,43 @@
...
+    SELECT array_agg(ROW (c.fact_id, c.version, c.path)::transformation_cache_entry_key)
+    INTO keys
+    FROM transformation_cache c
+    WHERE c.header ->> 'ns' = p_ns
+      AND c.header ->> 'type' = p_type
+      AND (
+        -- any of the version is contained in the path
+        (c.path && ARRAY [p_from_version, p_to_version])
+            -- or not all versions are provided
+            OR p_from_version IS NULL OR p_to_version IS NULL);
```
mamup11, line 27: This is dangerous, no? if we pass null then it will match all ns and type, no filter on the version, effectively what we had before? 

I guess this data should always be present, so we should not account for nulls here

factcast-store/src/main/resources/db/changelog/factcast/issue4872/notify_transformationstore_change.sql
```diff
@@ -0,0 +1,21 @@
+DROP TRIGGER IF EXISTS tr_deferred_transformationstore_change ON transformationstore;
```
mamup11, line 1: This drops the old trigger tr_deferred_transformationstore_change, meaning the old function that had a typo `notifyTrasformationStoreChange` is now orphaned, should we deleted on this liquidbase update?

factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TYPE transformation_cache_entry_key AS
```
mamup11, line 1: IF NOT EXISTS?

factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
```diff
@@ -0,0 +1,43 @@
...
+(
+    fact_id uuid,
+    version int,
+    path    int[]
+);
+
+CREATE PROCEDURE invalidate_transformation_cache(
```
mamup11: OR REPLACE

factcast-store/src/main/java/org/factcast/store/internal/notification/TransformationStoreChangeNotification.java
```diff
@@ -31,17 +31,40 @@ public class TransformationStoreChangeNotification extends StoreNotification {
   @NonNull String ns;
   @NonNull String type;
   long txId;
+  Integer fromVersion;
+  Integer toVersion;
+
+  public TransformationStoreChangeNotification(String ns, String type, long txId) {
```
mamup11: Missing @NonNull like the method bellow?

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
```diff
@@ -99,6 +99,23 @@ public void invalidateTransformationFor(@Nonnull String ns, @Nonnull String type
     }
   }
 
+  @Override
+  public void invalidateTransformationFor(String ns, String type, int fromVersion, int toVersion) {
```
mamup11: Also @nonNull?

factcast-store/src/main/resources/db/changelog/factcast/issue4872/notify_transformationstore_change.sql
```diff
@@ -0,0 +1,21 @@
...
+                                                    )::text);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE CONSTRAINT TRIGGER tr_deferred_transformationstore_change
+    AFTER DELETE OR UPDATE
+    ON transformationstore DEFERRABLE INITIALLY DEFERRED
+    FOR EACH ROW
+EXECUTE FUNCTION notifyTransformationStoreChange();
```
mamup11, line 23: Analyze is needed in the end so the planner gets the new index right?

factcast-store/src/test/java/org/factcast/store/registry/transformation/cache/PgTransformationCacheIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+class PgTransformationCacheIntegrationTest {
+  @Autowired JdbcTemplate jdbc;
+  @Autowired DataSource dataSource;
+  @Autowired PlatformTransactionManager transactionManager;
+
+  private PgTransformationCache uut;
+
+  @BeforeEach
+  void createCache() {
+    jdbc.execute("TRUNCATE transformation_cache");
```
mamup11: This should already be taken care of by the wipe.sql

factcast-store/src/test/java/org/factcast/store/registry/transformation/cache/PgTransformationCacheIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+            .extracting(TransformationStoreChangeNotification::uniqueId)
+            .doesNotHaveDuplicates();
+      } finally {
+        statement.execute("UNLISTEN transformationstore_change");
+      }
+    }
+  }
+
+  @Test
+  void noMatchingKeysDoesNotWaitForExclusiveLock() throws Exception {
```
mamup11, line 163: I dont fully get this test, what are you trying to test here? Also I saw no assertions so I was confused as to what is the indication of pass, just no exceptions?

factcast-store/src/test/java/org/factcast/store/registry/transformation/cache/PgTransformationCacheIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+    // Legacy notifications still invalidate all paths for the namespace and type.
+    uut.invalidateTransformationFor("ns", "type");
+
+    survivors.forEach(key -> assertThat(uut.find(key)).isEmpty());
+    assertThat(uut.find(otherNsKey)).isPresent();
+    assertThat(uut.find(otherTypeKey)).isPresent();
+  }
+
+  @Test
+  void createsNamespaceAndTypeIndex() {
```
mamup11: This test I dont think is needed? is basically testing liquidbase right? if the index exists. I dont think we need to now test this manually. But if you think otherwise thats ok

factcast-store/src/main/resources/db/changelog/factcast/issue4872/notify_transformationstore_change.sql
mamup11, line 1: I dont really like to have "business logic" on the DB, something we could do on the java side exactly the same: Query -> then with lock delete

The only reason I think this may be worth is keep is to avoid sending the whole key through the network, less data flying, but for something that happens not too often (or not at all). I think having this business logic here will lead to everybody forgetting this is here, and maybe make issues with it harder to find and triage

mamup11: Plus java side has better logging for our monitoring and alarms :D

factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
uweschaefer: transition: if the database was not updated yet. If we receive an old notification, yes, we react as we did before.

factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
uweschaefer: There is no if not exists or replace on create type
https://www.postgresql.org/docs/current/sql-createtype.html

factcast-store/src/main/java/org/factcast/store/internal/notification/TransformationStoreChangeNotification.java
uweschaefer: for ns & type, yes.

factcast-store/src/main/java/org/factcast/store/registry/transformation/cache/InMemTransformationCache.java
uweschaefer: can be null due to Object.equals and yodacoding. But i guess they should not.

factcast-store/src/main/resources/db/changelog/factcast/issue4872/invalidate_transformation_cache.sql
uweschaefer: also the notification could come via redis for instance.

factcast-store/src/main/resources/db/changelog/factcast/issue4872/notify_transformationstore_change.sql
uweschaefer: Having a hard time understand what "business logic" you refer to!?

factcast-store/src/test/java/org/factcast/store/registry/transformation/cache/PgTransformationCacheIntegrationTest.java
uweschaefer: it uses two connections, one locking the table, the other calling the procedure. if the procedure would try to lock, it would block forever. The 2s timeout on the statement would make this apparent by throwing an exception.

PR #4972, opened 2026-09-17
#3904: avoid backwards index scan when creating or validating statetoken

### LINKED ISSUE #3904: Performance bug: create state token

uweschaefer:
Here's what robo-friend says:

Previously, PgQueryBuilder.createStateSQL() generated a query structured as:

SELECT ser FROM fact
WHERE <whereClause> AND ser > ?
ORDER BY ser DESC LIMIT 1

When evaluating this query, the PostgreSQL cost-based optimizer makes decisions based on the combination of ORDER BY ser DESC and LIMIT 1:

1. Planner Heuristic on LIMIT 1: The PostgreSQL planner assumes data matching arbitrary predicates (such as header @> ?::jsonb) is uniformly distributed across the table. Because of LIMIT 1, the planner estimates that scanning the primary key
B-tree index (fact_pkey) backwards starting from the highest ser will find a matching row almost immediately.
2. Pathological Backward Index Scan: In reality, matching facts are often sparse, located far from the stream's tail, or non-existent (e.g., querying for a new aggregate or checking if new events arrived since the last token when none have).
3. Severe Degradation: Under the backward fact_pkey scan, PostgreSQL inspects millions of non-matching index entries and fetches heap pages backwards tuple by tuple until it hits a match or reaches ser > ?. This causes massive random I/O, buffer
cache thrashing, and latencies stretching into seconds or tens of seconds.

---

Why a Simple MAX(ser) Query Is Not Enough
One might consider replacing ORDER BY ser DESC LIMIT 1 with a direct aggregate:

SELECT MAX(ser) FROM fact WHERE <whereClause> AND ser > ?

However, the PostgreSQL optimizer recognizes SELECT MAX(col) on indexed columns and automatically optimizes it into an index scan on fact_pkey to locate the single maximum value matching the predicate. Thus, the planner falls back into the exact
same trap: a backward B-tree index scan over fact_pkey.

---

How the Materialized CTE Resolves the Issue
The refactored query structure:

WITH subq AS MATERIALIZED (
  SELECT ser FROM fact WHERE <whereClause>
)
SELECT COALESCE(MAX(ser), 0) FROM subq

solves the problem through several distinct database engine mechanics:

1. AS MATERIALIZED as an Optimization Fence
In PostgreSQL 12 and newer, WITH ... AS MATERIALIZED acts as a strict optimization fence. It instructs the query planner to evaluate the common table expression (CTE) independently as a subquery and materialize its intermediate result set in
memory, rather than inlining or folding it into the outer query.

Because of the fence:

• The outer aggregate MAX(ser) cannot be pushed down into the fact table scan.
• Outer clauses cannot bias the planner into attempting an index scan for min/max optimization.

2. Forcing GIN and Tail Index Usage Inside subq
Inside subq:

SELECT ser FROM fact WHERE <whereClause>

• There is no ORDER BY ser.
• There is no LIMIT 1.
• There is no aggregate function.

Because subq is a plain selection with no ordering incentives, PostgreSQL chooses the optimal access path for filtering JSONB predicates: a Bitmap Index Scan using the GIN index (idx_fact_header) or partial tail indexes (idx_fact_tail_*).

The GIN index directly identifies the exact row identifiers (TIDs) that satisfy the criteria without traversing non-matching entries:

• If no facts match, the index scan immediately reports zero matches in sub-millisecond time.
• If matching facts exist, only matching rows are fetched via a Bitmap Heap Scan.

3. Isolated Aggregation in the Outer Query
The outer query:

SELECT COALESCE(MAX(ser), 0) FROM subq

simply iterates over the small materialized intermediate table produced by subq and computes the maximum ser. Because subq only contains the filtered subset of matching facts, calculating MAX(ser) in memory takes negligible time.

4. Safe Default with COALESCE
If no rows match the filter, MAX(ser) over an empty set yields NULL. Wrapping the expression in COALESCE(MAX(ser), 0) guarantees:

• The query always returns a single row containing 0 instead of NULL or an empty result set.
• Existing extraction logic (ResultSetExtractor in PgFactStore) reliably receives 0L when no matching facts exist.

---

Execution Plan Comparison
Before (ORDER BY ser DESC LIMIT 1)
Limit (cost=... rows=1)
  -> Index Scan Backward using fact_pkey on fact
       Index Cond: (ser > ?)
       Filter: (header @> ?::jsonb)   <-- Evaluated row-by-row backwards for millions of rows!

After (WITH subq AS MATERIALIZED (...))
Aggregate (cost=...)
  -> CTE Scan on subq
       CTE subq
         -> Bitmap Heap Scan on fact
              Recheck Cond: (header @> ?::jsonb)
              Filter: (ser > ?)
              -> Bitmap Index Scan on idx_fact_header (or idx_fact_tail_*)
                   Index Cond: (header @> ?::jsonb)  <-- Direct index lookup, skips non-matching rows!

By decoupling the predicate evaluation from the MAX(ser) aggregation via the materialized CTE, the query replaces full-table backward scans with fast index lookups, ensuring predictable millisecond performance regardless of stream size.


uweschaefer:
closes #3904

factcast-site/documentation-docsy/content/en/Setup/properties.md
```diff
@@ -70,6 +70,7 @@ description: Properties you can use to configure FactCast
...
 | factcast.store.offload.name                                         | same as any DataSource                                                                                                                                                                                                                                                                                                                                                                                                                              |                                          |
+| factcast.store.state-query-backward-scan-window                     | amount of max rows to filter in a backwards scan to find the latest matching one for creating a statetoken, before falling back to a safer query method using the GIN indexes for filtering (slower for recent changes, much faster for very rare changes or no changes at all).                                                                                                                                                                    | 20000                                    |
```
leflamm: since today I was reminded that there might be gaps in the used serials 😛  ... 

```suggestion
| factcast.store.state-query-backward-scan-window                     | maximum number of serial positions to inspect in a backwards scan when creating or validating a state token, before falling back to a safer query method using the GIN indexes for filtering (slower for recent changes, much faster for very rare changes or no changes at all). Set to zero to disable the backwards probe..                                                                                                                                                                    | 20000                                    |
```

Last disable part only, if you want to state this so explicitly (`cutoff := max(ser)` and `ser > cutoff` will never match)

uweschaefer: validating has a (near) serial to start from. i'll remove the 0 bit, thanks for reminding me.

uweschaefer: and yes, it is not about rows but serials, but technically max rows is still valid and the difference is maybe more confusing than helping?

leflamm: fair point!

leflamm (approved):

leflamm:
LGTM! Thx :slightly_smiling_face:

PR #4974, opened 2026-09-17
#4973: make sure NudgeNotifications from txId 0 are not deuped

### LINKED ISSUE #4973: make sure NudgeNotifications will not be discarded as duplicates
**Describe the bug**
PgListener:158 must not use 0 as the txid for a nudgenotification, as they
might be deduplicated.

 post(new NudgeNotification(System.currentTimeMillis())); 

probably is enough unless we want a subclass for this kind of nudges

uweschaefer:


leflamm (approved):
LGTM. Thx!

PR #4977, opened 2026-09-17
#4976: Horizon 

### LINKED ISSUE #4976: reevaluate the concept of horizon or checkpoint
In order to close some gaps in highly concurrent unconditional publishing
situations, we want to take another attempt of testing the concept of a
checkpoint until which reading is safe (after that there might be
uncommitted inserts in flight), formerly known as 'horizon'.

uweschaefer:


factcast-store/src/main/java/org/factcast/store/internal/PgFactStream.java
```diff
@@ -271,24 +272,26 @@ void doCatchup() throws CatchupException {
...
+  private long executePhaseOne(PrimaryDataSourceSupplier primary, long primaryHorizonSerial)
+      throws CatchupException {
     if (offloadDataSource != null) {
       // we're creating a SCDS for offload, that we destroy right after
       try (SingleConnectionDataSource secondary =
           createCatchupDataSource(offloadDataSource, pipeline)) {
-        return catchupPhaseOne(secondary);
-      } catch (SQLException | PipelineAlreadyClosedException e) {
+        long offloadHorizonSerial =
+            Math.min(
```
otbe, line 291: Can we not just take always the HWM from the secondary? In my mind it should be equal or smaller than the primary one.

uweschaefer: don't think so. the primary was read some time before. if now inserts, advance and replication are fast enough, we could read beyond of what we're ffwding to. better safe than sorry.

uweschaefer: i'll add a comment though

